PR #21, opened 2018-04-09
Typographical -> spelling in README

theindigamer:
Minor fix: typographic means relating to [typography](https://en.wikipedia.org/wiki/Typography) which isn't related to spelling but rather to the layout of text and symbols.


theindigamer:
Oops, my bad, apparently Wikipedia does expand typo to typographical error.

Issue #25, opened 2018-04-21
Highlight matched query text in string[] fields


kishorenc:
## Description

Currently, matched text from a query is highlight only for plain `string` fields. For example, if `title` is a `string` field, and someone queries for `harry`, the result will be highlighted as follows:

```
"highlight": {
  "title": "<mark>Harry</mark> Potter and the Goblet of Fire"
}
```

The same should extend to a field of type `string[]`.

## Steps to reproduce

1. Index some text into `string[]` fields.
2. Try searching for some text against the `string[]` field.
3. Observe the `highlight` field in the response.

## Expected Behavior

On querying for `rowling` on an authors `string[]` field, highlighting of matched query text should be supported:

```
"highlight": {
  "authors": "J K <mark>Rowling</mark>"
}
```

## Actual Behavior

```
"highlight": { }
```

## Metadata

**Typsense Version**: `0.8.0`

**OS**: `all`

PR #26, opened 2018-05-05
Highlight for string array

kishorenc:
Fixes https://github.com/typesense/typesense/issues/25

This needs a change in the structure of the `highlight` field in response to allow sending the matched array index. For e.g. if a document has a `colors` array with a value of `["green", "blue", "red"]` and the search query is `red`, we need to communicate that matched array index is `2`:

```
...

"highlight": {
  "field": "colors",
  "snippet": "<mark>red</mark">,
  "index": 2
}
```

The `index` property will be present only a `string[]` field.

This PR also contains fixes for 2 other edge cases that should be part of the `0.9.0` change log:

1. A bug in fuzzy search involving prefixes - the prefix was being compared with the entire string for fuzzy matching instead of comparing with only the length of the overlapping segment.
2. When multiple fields were used for searching, in some cases, the same document was being returned multiple times leading to duplicate results.

### DIFF
--- a/docker/development.Dockerfile
+++ b/docker/development.Dockerfile
@@ -2,7 +2,7 @@
 # 
 # $ docker push typesense/typesense-development:latest
 
-FROM typesense/ubuntu-10.04-gcc:6.5.0
+FROM typesense/ubuntu-10-04-gcc:6.4.0
 
 ENV PATH /usr/local/gcc-6.4.0/bin/:$PATH
 ENV LD_LIBRARY_PATH /usr/local/gcc-6.4.0/lib64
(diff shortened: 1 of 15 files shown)


docker/development.Dockerfile
```diff
@@ -2,7 +2,7 @@
 # 
 # $ docker push typesense/typesense-development:latest
 
-FROM typesense/ubuntu-10.04-gcc:6.5.0
+FROM typesense/ubuntu-10-04-gcc:6.4.0
```
jasonbosco: Any reason this had to be downgraded?

docker/development.Dockerfile
```diff
@@ -16,26 +16,27 @@ RUN apt-get install -y python-software-properties \
...
-ADD https://cmake.org/files/v3.3/cmake-3.3.2-Linux-x86_64.tar.gz /opt/
-RUN tar -C /opt -xvzf /opt/cmake-3.3.2-Linux-x86_64.tar.gz
-RUN cp -r /opt/cmake-3.3.2-Linux-x86_64/* /usr
-
-ADD https://launchpad.net/ubuntu/+archive/primary/+files/snappy_1.1.3.orig.tar.gz /opt/
-RUN tar -C /opt -xf /opt/snappy_1.1.3.orig.tar.gz
-RUN mkdir /opt/snappy-1.1.3/build && cd /opt/snappy-1.1.3/build && ../configure && make && make install
-
-ADD https://openssl.org/source/openssl-1.0.2k.tar.gz /opt/
+RUN curl -L -o /opt/openssl-1.0.2k.tar.gz https://openssl.org/source/openssl-1.0.2k.tar.gz
```
jasonbosco: Any reason this can't be an `ADD` (vs `RUN`)? When using `ADD` it uses the host machine's internet connection and copies the files into the container. That way we don't need to have `curl` installed inside the docker container.

docker/development.Dockerfile
```diff
@@ -16,26 +16,27 @@ RUN apt-get install -y python-software-properties \
...
 RUN tar -C /opt -xvzf /opt/openssl-1.0.2k.tar.gz
 RUN cd /opt/openssl-1.0.2k && sh ./config --prefix=/usr --openssldir=/usr zlib-dynamic
 RUN make -C /opt/openssl-1.0.2k depend
 RUN make -C /opt/openssl-1.0.2k -j4
 RUN make -C /opt/openssl-1.0.2k install
 
-ADD https://github.com/curl/curl/releases/download/curl-7_55_1/curl-7.55.1.tar.bz2 /opt/
+RUN curl -L -o /opt/curl-7.55.1.tar.bz2 https://github.com/curl/curl/releases/download/curl-7_55_1/curl-7.55.1.tar.bz2
```
jasonbosco: Any reason this can't be an `ADD` (vs `RUN`)? When using `ADD` it uses the host machine's internet connection and copies the files into the container. That way we don't need to have `curl` installed inside the docker container.

include/string_utils.h
```diff
@@ -9,14 +9,14 @@
 
 struct StringUtils {
     UErrorCode status;
-    icu::Transliterator* transliterator;
+    //icu::Transliterator* transliterator;
 
-    StringUtils():status(U_ZERO_ERROR),
-                  transliterator(icu::Transliterator::createInstance("Latin-ASCII", UTRANS_FORWARD, status)) {
+    StringUtils(): status(U_ZERO_ERROR) {
+        // transliterator(icu::Transliterator::createInstance("Latin-ASCII", UTRANS_FORWARD, status))
```
jasonbosco: Did you intend to leave this line commented out?

include/string_utils.h
```diff
@@ -9,14 +9,14 @@
...
     }
 
     ~StringUtils() {
-        delete transliterator;
+        //delete transliterator;
```
jasonbosco: Did you intend to leave this line commented out?

src/art.cpp
```diff
@@ -1359,10 +1372,10 @@ int art_fuzzy_search(art_tree *t, const unsigned char *term, const int term_len,
         std::sort(nodes.begin(), nodes.end(), compare_art_node_score);
     }
 
-    long long int time_micro = microseconds(std::chrono::high_resolution_clock::now() - begin).count();
+    //long long int time_micro = microseconds(std::chrono::high_resolution_clock::now() - begin).count();
```
jasonbosco: Did you intend to leave this line commented out?

docker/development.Dockerfile
kishorenc: I had initially tagged it as `6.4.0` to refer to the GCC version we were using. However, when you made some changes and republished with `6.5.0` it no longer referred to the gcc version! So, I have moved it back to `6.4.0`.

docker/development.Dockerfile
kishorenc: The problem with `ADD` is that it does not cache the URL fetches: https://github.com/moby/moby/issues/880

So all download steps run again and again when you want to do a build and test. It's recommended that we use `curl` if we want to cache downloads.

include/string_utils.h
kishorenc: Yes, this was dead code, but I wanted to leave it around as it might be something that would be revisited in near future and I took some time to figure out that `Latin-ASCII` format.

include/string_utils.h
kishorenc: Yes

src/art.cpp
kishorenc: Yes, this must have been commented out already but it was not.

Issue #27, opened 2018-05-05
Ability to restrict fields returned


kishorenc:
## Description

Currently, the entire index document is returned in the search response. Sometimes, it's desirable to have only a specific set of fields returned for performance reasons. Ideally, a whitelist and blacklist of fields would be desirable. 

## Steps to reproduce

N/A

## Expected Behavior

N/A

## Actual Behavior

N/A

## Metadata

**Typsense Version**: `all`

**OS**: `all`

PR #28, opened 2018-05-08
Include exclude fields

kishorenc:
Fixes https://github.com/typesense/typesense/issues/27

Introduces two additional parameters to the search end-point: `include_fields` and `exclude_fields` that enabled customization of the document fields that are returned in the search result response.

@jasonbosco Use `typesense/typesense:include_exclude_fields` Docker image for testing this.


jasonbosco:
Works great!

Issue #613, opened 2022-06-09
Cannot get IDs in bulk import


RedSnail:
## Description

Cannot get ID's of documents when using bulk import, It only responds with the fact of success/failure.
Of course,I could generate an ID dynamically, but it can cause problems when there are multiple clients inserting data.

## Expected Behavior

Getting a created document in response would be nice (like in importing a single document) 

## Metadata

**0.23.0**:

**OS**:

PR #615, opened 2022-06-10
Option to return document ID and body in import response.

RedSnail:
## Change Summary
I've applied behaviour that I mentioned in #613 as optional, now you can get ID's of bulk-inserted documents if you specify "return_res" query parameter.

I've tested it and it's working.

## PR Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.iml
 external*
 build*
+test-build*
 cmake-build*
 cmake-build-debug
 cmake-build-release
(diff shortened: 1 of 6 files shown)


kishorenc:
@RedSnail 

Thank you for the PR! 

I like the idea of returning the `id` in the import response (as you've requested in https://github.com/typesense/typesense/issues/613), but I'm not sure about returning the full document -- the responses can get pretty large.  Can we stick to returning just the `id`?

RedSnail:
> @RedSnail
> 
> Thank you for the PR!
> 
> I like the idea of returning the `id` in the import response (as you've requested in #613), but I'm not sure about returning the full document -- the responses can get pretty large. Can we stick to returning just the `id`?

I can specify both behaviours with different params. I'm developing an async client and I need a full document in response, 'cause I don't cache input entries by default to avoid memory pressure

kishorenc:
👍 Adding both sounds good. Let's call the params as `return_id` and `return_document`.

kishorenc:
Looks good now! My last comment is about adding a test, if it's not too much of a hassle, along the lines of this `ImportDocumentsEmplace ` test: https://github.com/typesense/typesense/blob/main/test/collection_test.cpp#L1252

Just need to assert for the presence of both `id` and `doc` fields in the response when the respective flags are enabled.

kishorenc (approved):

kishorenc:
Merged, thank you for the contribution!

Issue #396, opened 2021-10-03
ANTLR4 grammar for filter expression


kishorenc:
## Description

Typesense currently supports a simple `filter_by` syntax that can do ANDs between different fields, but does not support ORs between fields or compound filter conditions. We want to support advance filtering operations and that requires generating a grammar for the existing filter syntax and extending it to support ORs (via `||`) and compound expressions. 

We've decided to use [ANTLR4](https://github.com/antlr/antlr4) as the parser generator, and are looking for help from the community to generate the ANTLR4 grammar. You can see [various examples here](https://github.com/antlr/grammars-v4).

Once we have the grammar, we can generate the parser and implement the filtering operations. Here are some examples of Typesense `filter_by` expressions that we can use as a reference for the grammar:

## Existing syntax

```
# Filtering a single value

name: Steve Smith
age: 10
is_enabled: true
```

```
# Strings can be wrapped by backticks to escape symbols that might conflict with filter syntax

name: `Toccata & Fugue`
```

```
# Filtering multiple values of the same field (we do an OR)

name: [Steve Smith, `Jack & Jill`]
age: [10, 15]
```

```
# Numerical comparators (we support <, >, <=, >= and ..)

age:>10
age: [>=10, 20]
age:[10..100]
```

```
# Not equals / Negation

author:!= JK Rowling
artist:!= [Michael Jackson, Taylor Swift]
```

```
# Exact equals (e.g. "shoes" in the example below will not match a field with the value "running shoes")

category:= shoes
category:= [`Running Shoes, Men`, Sneaker]
```

```
# ANDing multiple conditions

age:>20 && category:= [`Running Shoes, Men`, Sneaker]
```

```
# Geosearch related syntax

# Filtering around a radius (in either km or mi)
location:(48.906, 2.343, 5 mi)
location:(48.906, 2.343, 5 km)

# Within a polygon (can be any number of sides)
location:(48.8662, 2.3255, 48.8581, 2.3209, 48.8561, 2.3448, 48.8641, 2.3469)
```

## Additions

We now want to support compound filters and `||` (OR) conditions. Example:

```
((age: <5 || age: >10) && category:= [shoes]) || is_curated: true
```

PR #664, opened 2022-07-19
Filter query

happy-san:
## Change Summary
Work on #396 

## PR Checklist


include/field.h
```diff
@@ -433,6 +435,26 @@ struct filter {
...
+
+    static Option<bool> parse_filter_query2(const std::string& filter_query,
+                                            const std::unordered_map<std::string, field>& search_schema,
+                                            const Store* store,
+                                            const std::string& doc_id_prefix,
+                                            filter_node_t* root);
+};
+
+struct filter_node_t {
+    std::string token;
```
kishorenc, line 447: We can call this as `expression` since `token` means a word in the text in the rest of the project.

src/field.cpp
```diff
@@ -367,6 +369,416 @@ Option<bool> filter::parse_filter_query(const string& simple_filter_query,
...
+        } else if (token == "(") {
+            operatorStack.push(token);
+        } else if (token == ")") {
+            while (!operatorStack.empty() && operatorStack.top() != "(") {
+                postfix.push(operatorStack.top());
+                operatorStack.pop();
+            }
+
+            if (operatorStack.empty() || operatorStack.top() != "(") {
+                return Option<bool>(400, "Could not parse the filter query.");
```
kishorenc, line 401: We could probably given a more helpful message like `Could not parse the filter query: unbalanced parentheses.`

In other places also (e.g. in `toParseTree`), if we can give an explicit reasoning for parse error we should do so.

src/field.cpp
```diff
@@ -367,6 +369,416 @@ Option<bool> filter::parse_filter_query(const string& simple_filter_query,
...
+    auto token = node->token;
+    if(!isOperator(token)) {
+        // split into [field_name, value]
+        size_t found_index = token.find(':');
+
+        if(found_index == std::string::npos) {
+            return Option<bool>(400, "Could not parse the filter query.");
+        }
+
+        std::string&& field_name = token.substr(0, found_index);
```
kishorenc, line 471: `&&`?

src/field.cpp
```diff
@@ -367,6 +369,416 @@ Option<bool> filter::parse_filter_query(const string& simple_filter_query,
...
+            return Option<bool>(400, "Error with filter field `" + _field.name +
+                                     "`: Unidentified field data type, see docs for supported data types.");
+        }
+    }
+
+    return Option<bool>(true);
+}
+
+Option<bool> parse_filter_query2(const std::string& filter_query,
+                                        const std::unordered_map<std::string, field>& search_schema,
```
kishorenc, line 739: Indentation.

include/field.h
```diff
@@ -433,6 +435,26 @@ struct filter {
...
+    filter_node_t *left;
+    filter_node_t *right;
+    filter filter;
```
kishorenc, line 450: The build is not compiling because of naming the variable as the same as the type: 

```
error: declaration of 'filter filter_node_t::filter' changes meaning of 'filter'
```

We can call it as `filter_exp`.

kishorenc (changes requested):

src/field.cpp
happy-san: Didn't change the previous logic of creating a filter 
https://github.com/typesense/typesense/blob/dec1ed6533aeb04a74054020d2de603a52c334c6/src/field.cpp#L96

kishorenc (commented):
Looks good so far now. Can proceed with the `do_filtering2` implementation now. All tests should pass once that is done. We can also come back to add more unit tests to the parsing logic after that.

kishorenc:
Closing in favor of https://github.com/typesense/typesense/pull/737

PR #709, opened 2022-09-01
Added rate limiting feature

ozanarmagan:
## Change Summary
Added rate limiting feature with RateLimitManager singleton class. You can block, allow or throttle API keys or IP addresses.

## PR Checklist
- [ x ] I have read and signed the [Contributor License Agreement](https://forms.gle/PZyiY5N2GDQU8GsV9).

### DIFF
--- a/include/ratelimit_manager.h
+++ b/include/ratelimit_manager.h
@@ -175,7 +175,7 @@ class RateLimitManager
 
         // Prefix for bans
         static constexpr const char* BANS_NEXT_ID = "$KBN";
-        static constexpr const char* BANS_PREFIX = "$KB";
+        static constexpr const char* BANS_PREFIX = "$KBL";
 
 
 
--- a/include/ratelimit_manager.h
+++ b/include/ratelimit_manager.h
@@ -7,20 +7,10 @@
 #include <mutex>
 #include <shared_mutex>
 #include "lru/lru.hpp"
+#include "option.h"
 
 
-// Enum for rate limit ban durations from one minute to six hours
-enum class RateLimitBanDuration {
-    NO_BAN_BEFORE,
-    ONE_MINUTE,
-    FIVE_MINUTES,
-    TEN_MINUTES,
-    THIRTY_MINUTES,
-    ONE_HOUR,
-    THREE_HOURS,
-    SIX_HOURS,
-    PERMANENT
-};
+
 
 // Action enum for rate limit rules
 enum class RateLimitAction {
@@ -41,11 +31,13 @@ struct rate_limit_throttle_t {
 };
 
 struct rate_limit_rule_t {
-        uint64_t id;
-        RateLimitAction action;
-        RateLimitedResourceType resource_type;
-        std::vector<std::string> values;
-        rate_limit_throttle_t throttle;
+    uint64_t id;
+    RateLimitAction action;
+    RateLimitedResourceType resource_type;
+    std::vector<std::string> values;
+    rate_limit_throttle_t throttle;
+    int64_t auto_ban_threshold_num = -1;
+    int64_t auto_ban_num_days = -1;
 };
 
 // Entry struct for rate limit rule pointer hash map as key
@@ -79,25 +71,23 @@ struct request_counter_t {
 };
 
 // Struct to store ban information for IP addresses
-struct rate_limit_ban_t {
+struct rate_limit_status_t {
     bool is_banned = false;
     int64_t throttling_from;
     int64_t throttling_to;
     std::string value;
     RateLimitedResourceType resource_type;
-    RateLimitBanDuration banDuration = RateLimitBanDuration::NO_BAN_BEFORE;
 };
 
 
 // Hash function for rate_limit_rule_entry_t
-namespace std
-{
-template <>
-struct hash<rate_limit_rule_entry_t> {
-    std::size_t operator()(const rate_limit_rule_entry_t& c) const{
-        return std::hash<std::string>()(c.value);
-    }
-};
+namespace std {
+    template <>
+    struct hash<rate_limit_rule_entry_t> {
+        std::size_t operator()(const rate_limit_rule_entry_t& k) const{
+            return ((std::hash<int>()(static_cast<int>(k.resource_type)) ^ (std::hash<std::string>()(k.value) << 1)) >> 1);
+        }
+    };
 }
 
 
@@ -116,13 +106,13 @@ class RateLimitManager
         static RateLimitManager* getInstance();
         
         // Add rate limit for entries, returns false if there is a rule for any of the entries is already set
-        bool throttle_entries(const RateLimitedResourceType resource_type, const std::vector<std::string> &entries, const int64_t minute_rate_limit, const int64_t hour_rate_limit);
+        bool throttle_entries(const RateLimitedResourceType resource_type, const std::vector<std::string> &entries, const int64_t minute_rate_limit, const int64_t hour_rate_limit, const int64_t auto_ban_threshold_num, const int64_t auto_ban_num_days);
         
         // Remove rate limit for entry
         bool remove_rule_entry(const RateLimitedResourceType resource_type, const std::string &entry);
 
         // Get vector of banned entries
-        const std::vector<rate_limit_ban_t> get_banned_entries(const RateLimitedResourceType resource_type);
+        const std::vector<rate_limit_status_t> get_banned_entries(const RateLimitedResourceType resource_type);
 
         // Check if request is rate limited for given entries
         bool is_rate_limited(const std::vector<rate_limit_rule_entry_t> &entries);
@@ -134,7 +124,7 @@ class RateLimitManager
         bool allow_entries(RateLimitedResourceType resource_type, const std::vector<std::string> &entries);
 
         // Find rule by ID
-        const rate_limit_rule_t* find_rule_by_id(const uint64_t id);
+        Option<rate_limit_rule_t> find_rule_by_id(const uint64_t id);
 
         // Delete rule by ID
         bool delete_rule_by_id(const uint64_t id);
@@ -143,7 +133,7 @@ class RateLimitManager
         bool edit_rule_by_id(const uint64_t id, const rate_limit_rule_t& rule);
 
         // Get All rules 
-        const std::vector<rate_limit_rule_t>& get_all_rules();
+        const std::vector<rate_limit_rule_t> get_all_rules();
 
         // Clear all rules
         void clear_all();
@@ -158,22 +148,24 @@ class RateLimitManager
         inline static uint64_t last_rule_id = 0;
 
         // Store for rate_limit_rule_t
-        std::vector<rate_limit_rule_t> rule_store;
+        std::unordered_map<uint64_t,rate_limit_rule_t> rule_store;
 
         // LRU Cache to store rate limit and request counts for entries
         LRU::Cache<rate_limit_rule_entry_t, request_counter_t> rate_limit_request_counts;
+
         // Unordered map to point rules from rule store for entries
-        std::unordered_map<rate_limit_rule_entry_t, rate_limit_rule_t*> rate_limit_rule_pointer;
+        std::unordered_map<rate_limit_rule_entry_t, rate_limit_rule_t*> rate_limit_entries;
+        
         // Unordered map to store banned entries
-        std::unordered_map<rate_limit_rule_entry_t, rate_limit_ban_t> throttled_entries;
+        std::unordered_map<rate_limit_rule_entry_t, rate_limit_status_t> throttled_entries;
 
         // Mutex to protect access to ip_rate_limits and api_key_rate_limits
         std::shared_mutex rate_limit_mutex;
 
         // Helper function to ban an entry temporarily
-        void temp_ban_entry(const rate_limit_rule_entry_t& entry);
+        void temp_ban_entry(const rate_limit_rule_entry_t& entry, const int64_t number_of_days);
         // Helper function to ban an entry temporarily without locking mutex
-        void temp_ban_entry_wrapped(const rate_limit_rule_entry_t& entry);
+        void temp_ban_entry_wrapped(const rate_limit_rule_entry_t& entry, const int64_t number_of_days);
 
         // Singleton instance
         inline static RateLimitManager *instance;
--- a/include/ratelimit_manager.h
+++ b/include/ratelimit_manager.h
@@ -110,9 +110,6 @@ namespace std {
     };
 }
 
-class RateLimitManagerTest;
-
-
 class RateLimitManager 
 {
     public:
@@ -156,6 +153,9 @@ class RateLimitManager
 
         // Clear all rules
         void clear_all();
+
+        // Internal function to set base time
+        void _set_base_timestamp(const time_t& base_time);
 
         // Set store
         Option<bool> init(Store* store);
@@ -232,8 +232,6 @@ class RateLimitManager
 
         // Singleton instance
         inline static RateLimitManager *instance;
-
-        friend class RateLimitManagerTest;
 
 };
 
--- a/include/ratelimit_manager.h
+++ b/include/ratelimit_manager.h
@@ -2,6 +2,7 @@
 
 #include <string>
 #include <vector>
+#include <tuple>
 #include <unordered_map>
 #include <mutex>
 #include <shared_mutex>
@@ -21,47 +22,42 @@ enum class RateLimitBanDuration {
     PERMANENT
 };
 
-
-// Rate limit banned IP  structure with throttling_from and throttling_to fields.
-struct ratelimit_banned_ip_t {
-    std::string ip;
-    std::string api_key;
-    int64_t throttling_from;
-    int64_t throttling_to;
+// Action enum for rate limit rules
+enum class RateLimitAction {
+    ALLOW,
+    BLOCK,
+    THROTTLE
 };
 
+enum class RateLimitedResourceType {
+    IP,
+    API_KEY
+};
 
-
-
-// Rate limit struct which stores if IP or api_keey isTracked and rate limits for one minute and one hour.
-struct ratelimit_tracker_t {
-
-    private:
-
-        friend class RateLimitManager;
-
-        inline static int64_t last_id;
-    public:
-
-    ratelimit_tracker_t() {
-        id =  ratelimit_tracker_t::last_id++;
-    }
-
-    ratelimit_tracker_t(bool is_tracked, bool is_allowed, bool is_banned_permanently, int64_t minute_rate_limit, int64_t hour_rate_limit) :
-        is_tracked(is_tracked), is_allowed(is_allowed), is_banned_permanently(is_banned_permanently), minute_rate_limit(minute_rate_limit), hour_rate_limit(hour_rate_limit){
-        id = ratelimit_tracker_t::last_id++;
-    }
-    int64_t id;
-    bool is_tracked = false;
-    bool is_allowed = false;
-    bool is_banned_permanently = false;
+struct rate_limit_throttle_t {
     int64_t minute_rate_limit;
     int64_t hour_rate_limit;
-    std::string ip;
-    std::string api_key;
+
 };
 
+struct rate_limit_rule_t {
+        uint64_t id;
+        RateLimitAction action;
+        RateLimitedResourceType resource_type;
+        std::vector<std::string> values;
+        rate_limit_throttle_t throttle;
+};
 
+// Entry struct for rate limit rule pointer hash map as key
+struct rate_limit_rule_entry_t {
+    RateLimitedResourceType resource_type;
+    std::string value;
+
+    // Equality operator for rate_limit_rule_entry_t
+    bool operator==(const rate_limit_rule_entry_t& other) const {
+        return std::tie(resource_type, value) == std::tie(other.resource_type, other.value);
+    }
+};
 
 // Request counter struct for IP addresses to keep track of requests for current and previous sampling period
 struct request_counter_t {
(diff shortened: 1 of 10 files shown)


include/ratelimit_manager.h
```diff
@@ -0,0 +1,201 @@
...
+
+    // not-equal operator overload
+    bool operator!=(const request_counter_t& other) const{
+        return minute_rate_limit != other.minute_rate_limit || hour_rate_limit != other.hour_rate_limit || current_requests_count_minute != other.current_requests_count_minute || current_requests_count_hour != other.current_requests_count_hour || previous_requests_count_minute != other.previous_requests_count_minute || previous_requests_count_hour != other.previous_requests_count_hour || last_reset_time_minute != other.last_reset_time_minute || last_reset_time_hour != other.last_reset_time_hour;
+    }
+};
+
+// Struct to store ban information for IP addresses
+struct ratelimit_ban_t {
+    bool is_banned = false;
```
kishorenc, line 86: We should call this `is_throttled` to avoid confusion.

include/ratelimit_manager.h
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    ratelimit_tracker_t(bool is_tracked, bool is_allowed, bool is_banned_permanently, int64_t minute_rate_limit, int64_t hour_rate_limit) :
+        is_tracked(is_tracked), is_allowed(is_allowed), is_banned_permanently(is_banned_permanently), minute_rate_limit(minute_rate_limit), hour_rate_limit(hour_rate_limit){
+        id = ratelimit_tracker_t::last_id++;
+    }
+    int64_t id;
+    bool is_tracked = false;
+    bool is_allowed = false;
+    bool is_banned_permanently = false;
```
kishorenc, line 57: We should just call this `is_banned` and call the temporary ban flag as `is_throttled` (I've added comment for that below as well).

Also, should we just have an `action` enum here (like the JSON counterpart) rather than allow/ban etc. being individual boolean flags.

src/ratelimit_manager.cpp
```diff
@@ -0,0 +1,588 @@
...
+
+        // Check if current request count for hour rate limit is over the limit with sliding window
+        auto current_rate_hour = (3600 - (timeinfo -> tm_min * 60 + timeinfo -> tm_sec)) / 3600.0 * api_key_request_counts[api_key].previous_requests_count_hour + api_key_request_counts[api_key].current_requests_count_hour;
+        if (api_key_rate_limits[api_key].hour_rate_limit >= 0 && current_rate_hour >= api_key_rate_limits[api_key].hour_rate_limit) {
+            ban_api_key_wrapped(api_key);
+            return true;
+        }
+    }
+
+    // Check if IP is tracked
```
kishorenc, line 243: I see a lot of duplication in logic with the API key checks made above. This is making me think of just treating both of them as a singular entity. We can have a single hashmap, where the tracking identifier can can either be a IP address or an API key. The internal logic should not care what identifier is it. If we want we can prefix the key with "ip_" or "key_" to differentiate the entities. I feel that this will _vastly_ reduce the number of data structures we need to maintain and remove duplication of logic by making everything generic. What do you think?

kishorenc (changes requested):

kishorenc (changes requested):

src/ratelimit_manager.cpp
```diff
@@ -0,0 +1,338 @@
...
+        return;
+    }
+
+    auto now = std::chrono::system_clock::to_time_t(std::chrono::system_clock::now());
+    // Add entry to throttled_entries for the given number of days
+    throttled_entries.insert({entry, rate_limit_status_t{true, now, now + (number_of_days * 24 * 60 * 60), entry.value, entry.resource_type}});
+
+
+    if(rate_limit_request_counts.contains(entry)){
+        // Reset counters for the given entry
```
kishorenc, line 333: Since entry is throttled, we can just delete the entry from `rate_limit_request_counts`?

include/ratelimit_manager.h
```diff
@@ -0,0 +1,174 @@
...
+    int64_t minute_rate_limit;
+    int64_t hour_rate_limit;
+
+};
+
+struct rate_limit_rule_t {
+    uint64_t id;
+    RateLimitAction action;
+    RateLimitedResourceType resource_type;
+    std::vector<std::string> values;
```
kishorenc, line 37: Calling this as `entity_ids` is probably more clearer than `values` and goes well with `entity_type` above.

include/ratelimit_manager.h
```diff
@@ -0,0 +1,174 @@
...
+    rate_limit_throttle_t throttle;
```
kishorenc, line 38: In the same way, calling this `rate_limit_max_requests_t max_requests;` is more appropriate.

include/ratelimit_manager.h
```diff
@@ -0,0 +1,174 @@
...
+};
+
+enum class RateLimitedResourceType {
+    IP,
+    API_KEY
+};
+
+struct rate_limit_throttle_t {
+    int64_t minute_rate_limit;
+    int64_t hour_rate_limit;
```
kishorenc, line 29: Better to give this same name as the API:

```
struct rate_limit_max_requests_t {
    int64_t minute_threshold;
    int64_t hour_threshold;
}
```

include/ratelimit_manager.h
```diff
@@ -0,0 +1,174 @@
...
+    RateLimitAction action;
+    RateLimitedResourceType resource_type;
+    std::vector<std::string> values;
+    rate_limit_throttle_t throttle;
+    int64_t auto_ban_threshold_num = -1;
+    int64_t auto_ban_num_days = -1;
+};
+
+// Entry struct for rate limit rule pointer hash map as key
+struct rate_limit_rule_entry_t {
```
kishorenc, line 44: Let's call this `rate_limit_resource_t` as the `rule_entry` part makes it seem like it's an entry in the rule store, which it is not.

include/ratelimit_manager.h
```diff
@@ -0,0 +1,174 @@
...
+    RateLimitedResourceType resource_type;
+    std::string value;
```
kishorenc, line 46: This should be `entity_id`.

include/ratelimit_manager.h
```diff
@@ -0,0 +1,174 @@
...
+struct rate_limit_throttle_t {
+    int64_t minute_rate_limit;
+    int64_t hour_rate_limit;
+
+};
+
+struct rate_limit_rule_t {
+    uint64_t id;
+    RateLimitAction action;
+    RateLimitedResourceType resource_type;
```
kishorenc, line 36: Let's call this `RateLimitedEntityType entity_type`

include/ratelimit_manager.h
```diff
@@ -0,0 +1,174 @@
...
+
+
+// Action enum for rate limit rules
+enum class RateLimitAction {
+    ALLOW,
+    BLOCK,
+    THROTTLE
+};
+
+enum class RateLimitedResourceType {
```
kishorenc, line 22: Let's call this `RateLimitedEntityType`.

include/ratelimit_manager.h
```diff
@@ -0,0 +1,174 @@
...
+               std::tie(other.minute_rate_limit, other.hour_rate_limit, other.current_requests_count_minute, other.current_requests_count_hour, other.previous_requests_count_minute, other.previous_requests_count_hour, other.last_reset_time_minute, other.last_reset_time_hour);
+    }
+};
+
+// Struct to store ban information for IP addresses
+struct rate_limit_status_t {
+    bool is_banned = false;
+    int64_t throttling_from;
+    int64_t throttling_to;
+    std::string value;
```
kishorenc, line 78: This should be `entity_id`

include/ratelimit_manager.h
```diff
@@ -0,0 +1,174 @@
...
+    RateLimitedResourceType resource_type;
```
kishorenc, line 79: `RateLimitedEntityType entity_type`

src/ratelimit_manager.cpp
```diff
@@ -0,0 +1,338 @@
...
+            request_counts.current_requests_count_minute++;
+            request_counts.current_requests_count_hour++;
+
+            // Check if request count is over the limit
+            auto current_rate_for_minute = (60 - (std::chrono::system_clock::to_time_t(std::chrono::system_clock::now()) - request_counts.last_reset_time_minute)) / 60  * request_counts.previous_requests_count_minute;
+            current_rate_for_minute += request_counts.current_requests_count_minute;
+
+            if(rule.throttle.minute_rate_limit >= 0 && current_rate_for_minute > rule.throttle.minute_rate_limit) {
+                if(rule.auto_ban_threshold_num >= 0 && rule.auto_ban_num_days >= 0) {
+                    temp_ban_entry_wrapped(entry, rule.auto_ban_num_days);
```
kishorenc, line 174: This seems wrong. We don't have a counter for the number of times an IP / API key has been throttled. Only if __that count__ crosses `auto_ban_threshold_num` we should ban it.

kishorenc (changes requested):

src/ratelimit_manager.cpp
```diff
@@ -0,0 +1,344 @@
...
+            return false;
+        }
+        else if(throttled_entities.count(entity) > 0 && throttled_entities.at(entity).is_banned) {
+            
+            // Check if ban duration is not over
+            if (throttled_entities.at(entity).throttling_to > std::chrono::system_clock::to_time_t(std::chrono::system_clock::now())) {
+                
+                // Delete request counts for the entity
+                if(rate_limit_request_counts.contains(entity)) {
+                    rate_limit_request_counts.erase(entity);
```
kishorenc, line 111: I think erase is not needed here. Since `rate_limit_request_counts` is an LRU, it should get cycled out.

src/ratelimit_manager.cpp
```diff
@@ -0,0 +1,344 @@
...
+                }
+
+                return true;
+            }
+ 
+            // Remove ban
+            throttled_entities.at(entity).is_banned = false;
+
+            // Check if throttle rule still exists
+            if (rate_limit_entities.count(entity) > 0 && (rate_limit_entities.at(entity)->action == RateLimitAction::ALLOW)) {
```
kishorenc, line 121: Already checking for `rate_limit_entities.count(entity)` above in line 92. We can maybe use a `.find` there so that the fetched value can be reused here.

include/ratelimit_manager.h
```diff
@@ -0,0 +1,175 @@
...
+    RateLimitAction action;
+    RateLimitedEntityType entity_type;
+    std::vector<std::string> entity_ids;
+    rate_limit_max_requests_t max_requests;
+    int64_t auto_ban_threshold_num = -1;
+    int64_t auto_ban_num_days = -1;
+};
+
+// Entry struct for rate limit rule pointer hash map as key
+struct rate_limit_resource_t {
```
kishorenc, line 44: This should be `rate_limit_entity_t`.

src/ratelimit_manager.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+
+bool RateLimitManager::is_rate_limited(const std::vector<rate_limit_resource_t> &entities) {
+    // lock mutex
+    std::unique_lock<std::shared_mutex>lock(rate_limit_mutex);
+    
+    // Check if any of the entities has rule
+    for(const auto &entity : entities) {
+
+        if (rate_limit_entities.count(entity) == 0) {
```
kishorenc, line 92: How will you handle wildcard `.*` rules here? For e.g. rule for any IP or any API key will use `.*` for IP address or API key.

src/ratelimit_manager.cpp
```diff
@@ -0,0 +1,344 @@
...
+    if(rule_store.count(id) > 0) {
+        auto rule = rule_store.at(id);
+
+        // Remove rule from rate limit rule pointer
+        for(const auto &entity : rule.entity_ids) {
+            rate_limit_entities.erase(rate_limit_resource_t{rule.entity_type,entity});
+        }
+
+        // Update rule in rule store
+        rule_store.at(id) = new_rule;
```
kishorenc, line 266: It's okay to use `[ ]` during assignment. Only don't use it for gets.

src/ratelimit_manager.cpp
```diff
@@ -0,0 +1,344 @@
...
+    }
+
+    return false;
+}
+
+bool RateLimitManager::edit_rule_by_id(const uint64_t id, const rate_limit_rule_t & new_rule) {
+    std::unique_lock<std::shared_mutex> lock(rate_limit_mutex);
+
+    // Check if a rule exists for the given ID
+    if(rule_store.count(id) > 0) {
```
kishorenc, line 257: Use `rule_store.find()` when you want to do something with the found object. Better than doing a `count` and then an `at` later on.

src/ratelimit_manager.cpp
```diff
@@ -0,0 +1,344 @@
...
+                // Delete request counts for the entity
+                if(rate_limit_request_counts.contains(entity)) {
+                    rate_limit_request_counts.erase(entity);
+                }
+
+                return true;
+            }
+ 
+            // Remove ban
+            throttled_entities.at(entity).is_banned = false;
```
kishorenc, line 118: Why should we set a flag here instead of removing entity from the `throttled_entities` map? In fact, I think the `is_banned` flag is redundant here because a presence of an entity in this map already means it's throttled.

kishorenc (changes requested):

src/core_api.cpp
```diff
@@ -1693,3 +1695,244 @@ bool del_preset(const std::shared_ptr<http_req>& req, const std::shared_ptr<http
...
+    return true;
+}
+
+bool del_rate_limit(const std::shared_ptr<http_req>& req, const std::shared_ptr<http_res>& res) {
+    RateLimitManager* rateLimitManager = RateLimitManager::getInstance();
+    uint64_t id = std::stoull(req->params["id"]);
+    const auto& ruleOption = rateLimitManager->find_rule_by_id(id);
+
+    if(!ruleOption.ok()) {
+        res->set_404();
```
kishorenc, line 1817: `ruleOption` could in future return non-404 errors as well. Better to do:

```
res->set(ruleOption.code(), ruleOption.error());
```

Check other places where this has been done and fix those as well.

src/core_api.cpp
```diff
@@ -1693,3 +1695,244 @@ bool del_preset(const std::shared_ptr<http_req>& req, const std::shared_ptr<http
...
+        res->set_400("Bad JSON.");
+        return false;
+    }
+
+    if (req_json.count("action") == 0) {
+        res->set_400("Parameter `action` is required.");
+        return false;
+    }
+
+    if (req_json["action"] == "allow") {
```
kishorenc, line 1842: All the processing below is duplicated with the `put_rate_limit` above. We should probably extract them out and maybe make them as part of rateLimitManager class.

src/ratelimit_manager.cpp
```diff
@@ -0,0 +1,388 @@
...
+    if(rate_limit_request_counts.contains(entity)){
+        // Reset counters for the given entity
+        rate_limit_request_counts.lookup(entity).current_requests_count_minute = 0;
+        rate_limit_request_counts.lookup(entity).current_requests_count_hour = 0;
+    }
+
+}
+
+const nlohmann::json RateLimitManager::get_all_rules_json()
+{
```
kishorenc, line 351: Curly must be on the previous line.

src/ratelimit_manager.cpp
```diff
@@ -0,0 +1,388 @@
...
+
+const nlohmann::json rate_limit_rule_t::to_json() const {
+        nlohmann::json rule;
+        rule["id"] = id;
+        rule["action"] = magic_enum::enum_name(action);
+        rule["entity_type"] = magic_enum::enum_name(entity_type);
+        rule["entity_ids"] = entity_ids;
+        
+        if(max_requests.minute_threshold >= 0) {
+        rule["max_requests"]["minute_threshold"] = max_requests.minute_threshold;
```
kishorenc, line 372: Indentation. Check other `if` below as well.

test/ratelimit_test.cpp
```diff
@@ -0,0 +1,224 @@
...
+    EXPECT_TRUE(manager->get_all_rules().size() == 2);
+}
+
+TEST_F(RateLimitManagerTest, TestGetAllRulesEmpty) {
+    auto rules = manager->get_all_rules();
+    EXPECT_EQ(rules.size(), 0);
+}
+
+TEST_F(RateLimitManagerTest, TestGetAllRulesJSON)
+{
```
kishorenc, line 215: Curly should be on the previous line.

test/ratelimit_test.cpp
```diff
@@ -0,0 +1,224 @@
...
+#include "ratelimit_manager.h"
+
+// Google test for RateLimitManager
+class RateLimitManagerTest : public ::testing::Test
+{
+protected:
+    RateLimitManager *manager = RateLimitManager::getInstance();
+
+    RateLimitManagerTest()
+    {
```
kishorenc, line 13: Formatting of `{` and `}`.

src/ratelimit_manager.cpp
```diff
@@ -0,0 +1,388 @@
...
+            }
+
+
+
+            // Check if request count is over the limit
+            auto current_rate_for_minute = (60 - (std::chrono::system_clock::to_time_t(std::chrono::system_clock::now()) - request_counts.last_reset_time_minute)) / 60  * request_counts.previous_requests_count_minute;
+            current_rate_for_minute += request_counts.current_requests_count_minute;
+
+            if(rule.max_requests.minute_threshold >= 0 && current_rate_for_minute >= rule.max_requests.minute_threshold) {
+                if(rule.auto_ban_threshold_num >= 0 && rule.auto_ban_num_days >= 0) {
```
kishorenc, line 172: `auto_ban_threshold_num` must actually be greater than 0, because `auto_ban_threshold_num: 0` has no meaning. Likewise for `auto_ban_num_days`.

Also, can we extract this into a meaningful boolean variable like this?

```
bool auto_ban_is_enabled = (rule.auto_ban_threshold_num > 0 && rule.auto_ban_num_days > 0);
```

Helps with readability.

src/ratelimit_manager.cpp
```diff
@@ -0,0 +1,388 @@
...
+                    if(++request_counts.threshold_exceed_count_minute > rule.auto_ban_threshold_num) {
```
kishorenc, line 173: The `++request_counts.threshold_exceed_count_minute` increment is easy to miss when it's inline. Have it on a separate line:

```
request_counts.threshold_exceed_count_minute++;
```

kishorenc (changes requested):

src/ratelimit_manager.cpp
```diff
@@ -67,11 +68,19 @@ bool RateLimitManager::is_rate_limited(const std::vector<rate_limit_entity_t> &e
             if (throttled_entities.at(entity).throttling_to > std::chrono::system_clock::to_time_t(std::chrono::system_clock::now())) {
                 return true;
             }
+
+
+            // Remove ban from DB store
+            std::string ban_key = std::string(BANS_PREFIX) + "_" + std::to_string(throttled_entities.at(entity).status_id);
```
kishorenc, line 74: The ban key construction logic must be a helper method because it's duplicated in atleast one another place.

src/ratelimit_manager.cpp
```diff
@@ -314,13 +361,22 @@ Option<nlohmann::json> RateLimitManager::add_rule(const nlohmann::json &rule_jso
 
     parsed_rule.id = last_rule_id++;
 
-    rule_store[parsed_rule.id] = parsed_rule;
+    const std::string rule_store_key = std::string(RULES_PREFIX) + "_" + std::to_string(parsed_rule.id);
```
kishorenc, line 364: Rule store key construction should be a helper method.

include/ratelimit_manager.h
```diff
@@ -148,14 +155,36 @@ class RateLimitManager
...
             rate_limit_request_counts.capacity(10000);
         }
 
+        // Store for rate limit rules
+        Store *store;
+
+        // Using a $ prefix so that these meta keys stay above record entries in a lexicographically ordered KV store
+
+        // Prefix for rate limit rules
+        static constexpr const char* RULES_NEXT_ID = "$KRLN";
```
kishorenc, line 173: The `K` is superfluous. Can just begin everything with `$RL...` instead (RL for RateLimit component). So:

- $RLN
- $RLRP
- $RLBN
- $RLBP

src/ratelimit_manager.cpp
```diff
@@ -491,6 +553,97 @@ Option<rate_limit_rule_t> RateLimitManager::parse_rule(const nlohmann::json &rul
...
+    for(const auto& rule_json_str: rule_json_strs) {
+
+        nlohmann::json rule_json = nlohmann::json::parse(rule_json_str);
+
+        Option<rate_limit_rule_t> rule_option = parse_rule(rule_json);
+
+        if(!rule_option.ok()) {
+            return Option<bool>(rule_option.code(), rule_option.error());
+        }
+
```
kishorenc, line 603: Please check for extra white lines like these and remove them.

kishorenc (changes requested):

test/ratelimit_test.cpp
```diff
@@ -345,3 +350,45 @@ TEST_F(RateLimitManagerTest, TestGetAllRulesJSON) {
...
+        {"api_keys", nlohmann::json::array({".*"})},
+        {"max_requests_60s", 5},
+        {"max_requests_1h", -1}
+    });
+    EXPECT_FALSE(manager->is_rate_limited({{RateLimitedEntityType::api_key, "test"}}));
+    EXPECT_FALSE(manager->is_rate_limited({{RateLimitedEntityType::api_key, "test"}}));
+    EXPECT_FALSE(manager->is_rate_limited({{RateLimitedEntityType::api_key, "test"}}));
+    EXPECT_FALSE(manager->is_rate_limited({{RateLimitedEntityType::api_key, "test"}}));
+    EXPECT_FALSE(manager->is_rate_limited({{RateLimitedEntityType::api_key, "test"}}));
+    EXPECT_TRUE(manager->is_rate_limited({{RateLimitedEntityType::api_key, "test"}}));
```
kishorenc, line 393: 1. Should test the same against another key like `test1` to "prove" that `.*` works. 
2. I also want to see a test of a custom explicit API key rule along with `.*` -- which one will be picked first?

include/ratelimit_manager.h
```diff
@@ -222,10 +224,16 @@ class RateLimitManager
...
+        // Base timestamp
+        time_t base_timestamp = 0;
+
+        // Helper function to get current timestamp
+        time_t get_current_time();
 
         // Singleton instance
         inline static RateLimitManager *instance;
 
+        friend class RateLimitManagerTest;
```
kishorenc, line 236: Not a fan of friend class. Let's have a public `_set_current_time` function that the test uses (name begins with `_` to indicate internal use).  With this, we also don't need forward declaration of `RateLimitManagerTest` above.

test/ratelimit_test.cpp
```diff
@@ -345,3 +350,45 @@ TEST_F(RateLimitManagerTest, TestGetAllRulesJSON) {
     EXPECT_EQ(rules.at(0).at("entity_type").is_string(), true);
     EXPECT_EQ(rules.at(0).at("api_keys").is_array(), true);
 }
+
+TEST_F(RateLimitManagerTest, TestAutoBan) {
```
kishorenc, line 354: Not related to this test: do we have a test where we multiple rules that all match (e.g. both `allow` and `deny`) and ensure that the correct order is picked?

kishorenc (changes requested):

kishorenc (approved):
Thanks, this is great work!

PR #812, opened 2022-11-30
Rate limit update & fixes

ozanarmagan:
List of changes:

1.Added a more meaningful response for delete operation of rules (returns ID of the deleted rule as requested)
2./limits/active endpoint is active now.
3. You can delete auto bans with DELETE request to /limits/active/:id endpoint
4. Resolved hanging issue for /limits/:id endpoint with invalid inputs.
5./multi_search endpoint now increases request counts as size of searches array.
6.Encountered and fixed a logic error on calculation of how many times user exceeded minute threshold so far.
7.Added unit test for those cases.

## PR Checklist

### DIFF
--- a/include/ratelimit_manager.h
+++ b/include/ratelimit_manager.h
@@ -164,8 +164,13 @@ class RateLimitManager
         // Internal function to set base time
         void _set_base_timestamp(const time_t& base_time);
 
+        // Init
+        Option<bool> init();
+
         // Set store
-        Option<bool> init(Store* store);
+        inline void set_store(Store* store) {
+            this->store = store;
+        }
 
     private:    
 
--- a/include/ratelimit_manager.h
+++ b/include/ratelimit_manager.h
@@ -40,7 +40,7 @@ struct rate_limit_rule_t {
     std::vector<std::string> entity_ids;
     rate_limit_max_requests_t max_requests;
     int64_t auto_ban_threshold_num = -1;
-    int64_t auto_ban_num_days = -1;
+    int64_t auto_ban_num_hours = -1;
 
     const nlohmann::json to_json() const;
 
@@ -211,15 +211,15 @@ class RateLimitManager
         std::shared_mutex rate_limit_mutex;
 
         // Helper function to ban an entity temporarily
-        void temp_ban_entity(const rate_limit_entity_t& entity, const int64_t number_of_days);
+        void temp_ban_entity(const rate_limit_entity_t& entity, const int64_t number_of_hours);
         // Helper function to ban an entity temporarily without locking mutex
-        void temp_ban_entity_unsecure(const rate_limit_entity_t& entity, const int64_t number_of_days);
+        void temp_ban_entity_unsecure(const rate_limit_entity_t& entity, const int64_t number_of_hours);
 
         // Helper function to check if JSON rule is valid
         Option<bool> is_valid_rule(const nlohmann::json &rule_json);
 
         // Parse JSON rule to rate_limit_rule_t
-        Option<rate_limit_rule_t> parse_rule(const nlohmann::json &rule_json, bool alert_if_exists = true);
+        static rate_limit_rule_t parse_rule(const nlohmann::json &rule_json);
 
         // Helper function to insert rule in store
         void insert_rule(const rate_limit_rule_t &rule);
--- a/include/ratelimit_manager.h
+++ b/include/ratelimit_manager.h
@@ -165,14 +165,9 @@ class RateLimitManager
         void _set_base_timestamp(const time_t& base_time);
 
         // Init
-        Option<bool> init();
+        Option<bool> init(Store* store);
 
-        // Set store
-        inline void set_store(Store* store) {
-            this->store = store;
-        }
-
-    private:    
+    private:   
 
         RateLimitManager() {
             rate_limit_request_counts.capacity(10000);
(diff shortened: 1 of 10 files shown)


include/ratelimit_manager.h
```diff
@@ -151,14 +152,25 @@ class RateLimitManager
...
 
         // Internal function to set base time
         void _set_base_timestamp(const time_t& base_time);
 
+        // Init
+        Option<bool> init();
+
         // Set store
-        Option<bool> init(Store* store);
+        inline void set_store(Store* store) {
```
kishorenc, line 171: Setters like this have to be avoided. For singletons, better to use `init()` convention.

src/core_api.cpp
```diff
@@ -1703,6 +1716,13 @@ bool get_rate_limits(const std::shared_ptr<http_req>& req, const std::shared_ptr
 
 bool get_rate_limit(const std::shared_ptr<http_req>& req, const std::shared_ptr<http_res>& res) {
     RateLimitManager* rateLimitManager = RateLimitManager::getInstance();
+
+    // check if id is numeric
+    if(!StringUtils::is_int64_t(req->params["id"])) {
```
kishorenc, line 1721: For `stoull` we need to use `is_uint64_t` (unsigned version).

src/core_api.cpp
```diff
@@ -1718,6 +1738,13 @@ bool get_rate_limit(const std::shared_ptr<http_req>& req, const std::shared_ptr<
 
 bool put_rate_limit(const std::shared_ptr<http_req>& req, const std::shared_ptr<http_res>& res) {
     RateLimitManager* rateLimitManager = RateLimitManager::getInstance();
+
+    // check if id is numeric
+    if(!StringUtils::is_int64_t(req->params["id"])) {
```
kishorenc, line 1743: For `stoull` we need to use `is_uint64_t` (unsigned version).

src/core_api.cpp
```diff
@@ -1742,6 +1769,13 @@ bool put_rate_limit(const std::shared_ptr<http_req>& req, const std::shared_ptr<
 
 bool del_rate_limit(const std::shared_ptr<http_req>& req, const std::shared_ptr<http_res>& res) {
     RateLimitManager* rateLimitManager = RateLimitManager::getInstance();
+
+    // check if id is numeric
+    if(!StringUtils::is_int64_t(req->params["id"])) {
```
kishorenc, line 1774: For `stoull` we need to use `is_uint64_t` (unsigned version).

src/http_server.cpp
```diff
@@ -497,6 +497,11 @@ int HttpServer::catch_all_handler(h2o_handler_t *_h2o_handler, h2o_req_t *req) {
 
     //LOG(INFO) << "Init res: " << custom_gen->response << ", ref count: " << custom_gen->response.use_count();
 
+    if(rpath->action == "documents:search" && path_without_query == "/multi_search") {
```
kishorenc, line 500: Use `if(root_resource == "multi_search") {`.

src/http_server.cpp
```diff
@@ -497,6 +497,11 @@ int HttpServer::catch_all_handler(h2o_handler_t *_h2o_handler, h2o_req_t *req) {
...
+        // append api key and ip to metadata
+        request->metadata = api_auth_key_sent + " " + client_ip;
```
kishorenc, line 502: What if `api_auth_key_sent` contains a space (we don't enforce character check on the API key). We can instead use this format:

```
<length_of_api_key>:<api_key><client_ip>
```

We will read string until `:`, then extract the API key length, then extract the API key based on length and then finally the client_ip (rest of the string).

src/raft_server.cpp
```diff
@@ -499,6 +500,14 @@ int ReplicationState::init_db() {
         return 1;
     }
 
+    RateLimitManager *rateLimitManager = RateLimitManager::getInstance();
```
kishorenc, line 503: Since we are using the `meta_store` now, it's okay to init the rate limit manager in the earlier place. We should not do it here.

src/ratelimit_manager.cpp
```diff
@@ -165,18 +168,17 @@ const std::vector<rate_limit_rule_t> RateLimitManager::get_all_rules() {
 
 const std::vector<rate_limit_status_t> RateLimitManager::get_banned_entities(const RateLimitedEntityType entity_type) {
     std::shared_lock<std::shared_mutex> lock(rate_limit_mutex);
```
kishorenc, line 170: Existing convention is `std::shared_lock lock(rate_limit_mutex);` rather than `std::shared_lock<std::shared_mutex> lock(rate_limit_mutex)`. (less verbose also). Check other instances of this pattern to fix as well.

src/ratelimit_manager.cpp
```diff
@@ -491,4 +501,55 @@ time_t RateLimitManager::get_current_time() {
 
 void RateLimitManager::_set_base_timestamp(const time_t& timestamp) {
     base_timestamp = timestamp;
+}
+
+
+const nlohmann::json RateLimitManager::get_all_throttled_entities_json() {
+    nlohmann::json throttled_entities_array = nlohmann::json::object();
```
kishorenc, line 508: No lock? Also go through all other methods in rate limit manager to ensure that locks are not missed.

test/ratelimit_test.cpp
```diff
@@ -36,7 +37,8 @@ class RateLimitManagerTest : public ::testing::Test
         system(("rm -rf "+state_dir_path+" && mkdir -p "+state_dir_path).c_str());
 
         store = new Store(state_dir_path);
-        manager->init(store);
+        manager->set_store(store);
```
kishorenc, line 40: See comment regarding setter.

kishorenc (changes requested):

PR #810, opened 2022-11-29
Add project seldon to the showcase

skipjack:
## Change Summary

We've been building this tool for a while but just starting to open it up and post about it more publicly. Would love for anyone to try it out and give us some feedback! TypeSense has been working like charm, even in the places where we're probably abusing its intended use case 😁

## PR Checklist

- [x] I have read and signed the [Contributor License Agreement](https://forms.gle/PZyiY5N2GDQU8GsV9).

Issue #140, opened 2020-10-23
Help us build a showcase


jasonbosco:
We're looking to build a showcase to show off the different uses of Typesense in production. If you do use Typesense, we'd love to include you in the showcase. Please add to this file and send us a PR: [SHOWCASE.md](https://github.com/typesense/typesense/blob/master/SHOWCASE.md)

AlexFromStoripress:
We'll be using Typesense in our app (not yet live), see www.storipress.com! Awesome work guys :)

agentile:
You know you can always add us @jasonbosco :)

jasonbosco:
Thanks @agentile :) I've been meaning to email you asking if I can add you, will do so now!

echo-slam-jam:
I would like to self host typesense in my VM but I am lost in trying to sync my firestore data to my local VM collection :(

jasonbosco:
@echo-slam-jam I just noticed your comment here! I've responded to your issue here: https://github.com/typesense/firestore-typesense-search/issues/12. Let's continue the conversation there.

echo-slam-jam:
> 
> 
> @echo-slam-jam I just noticed your comment here! I've responded to your issue here: [typesense/firestore-typesense-search#12](https://github.com/typesense/firestore-typesense-search/issues/12). Let's continue the conversation there.

So recap of what I need to do so far:

- put my domain name on the firebase extension HOST field
- put my public IP on the **api-address**
-change port to 443

jasonbosco:
Could you respond in the other issue I linked above, so it’s all in one place?

echo-slam-jam:
Oh sorry I misread 
Let's continue the conversation there.
as
Let's continue the conversation here.

jasonbosco:
Mind deleting your comments here so we can keep this thread clean?

echo-slam-jam:
> 
> 
> Mind deleting your comments on this issue so we can keep this thread clean?

Sure

Alexander-Zierhut:
@jasonbosco Sure! Should I simply create a PR for that as soon as our implementation is done?

jasonbosco:
@Alexander-Zierhut Yes, a PR would be great!

hitiksaini:
Hey! I am using flutter for some project and after comparing various searching engines, I decided to move forward with typesense. 

But the package :  [typesense 0.1.1](https://pub.dev/packages/typesense) does not support **null safety**

Any updates?

jasonbosco:
[...]

PR #816, opened 2022-12-03
Add project seldon to the showcase

skipjack:
## Change Summary

We've been building this tool for a while but just starting to open it up and post about it more publicly. Would love for anyone to try it out and give us some feedback! TypeSense has been working like charm, even in the places where we're probably abusing its intended use case 😁

## PR Checklist


skipjack:
@jasonbosco @kishorenc btw you may want to revise the link in #140 to point to `main` instead of `master`, hat's how I ended up against `master` in #810 (see the bottom left)...

image

jasonbosco (approved):

jasonbosco:
Ahhhh, I was wondering why folks were opening a PR from master instead of main, even though main was set as the default branch. Thank you for you pointing that out @skipjack! I've fixed the link.

PR #819, opened 2022-12-04
Correct docker command in README.md

normal-coder:
## Change Summary

Correct docker command in README.md

## PR Checklist


jasonbosco:
@normal-coder The existing command should work without a space after `-v` as well. This is how I've always run Typesense. Does this not work for you?

normal-coder:
> @normal-coder The existing command should work without a space after `-v` as well. This is how I've always run Typesense. Does this not work for you?

It works, for the usage habits of most command line parameters, it may be more intuitive to bring spaces.

PR #1052, opened 2023-06-08
adding found documents in results when group_by is enabled

krunal1313:
## PR Checklist

### DIFF
--- a/src/collection.cpp
+++ b/src/collection.cpp
@@ -1384,7 +1384,7 @@ Option<nlohmann::json> Collection::search(std::string  raw_query,
     std::vector<std::vector<KV*>> raw_result_kvs;
     std::vector<std::vector<KV*>> override_result_kvs;
 
-    size_t total_found = 0;
+    size_t total_found = 0, total_groups_found = 0;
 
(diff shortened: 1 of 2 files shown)


src/collection.cpp
```diff
@@ -1557,10 +1557,10 @@ Option<nlohmann::json> Collection::search(std::string  raw_query,
...
     // for grouping we have to aggregate group set sizes to a count value
     if(group_limit) {
-        total_found = search_params->groups_processed.size() + override_result_kvs.size();
-    } else {
-        total_found = search_params->all_result_ids_len;
-    }
+        total_groups_found = search_params->groups_processed.size() + override_result_kvs.size();
+    } 
+    
+    total_found = search_params->all_result_ids_len;
```
kishorenc, line 1563: We can't break existing behavior. `total_found` should remain as the current logic, i.e. number of groups for group queries and number of docs otherwise. Instead lets add `found_docs` as an additional field available for group queries.

kishorenc (changes requested):

kishorenc (approved):

PR #1097, opened 2023-07-10
Install Curl for health check support

nitrag:
## Change Summary
If you can install curl in the deployment image this will allow docker to monitor the health of the container:

```
healthcheck:
    test: [ "CMD", "curl", "-f", "http://localhost:8108/health" ]
    interval: 30s
    timeout: 5s
    retries: 1
    start_period: 30s
```

## PR Checklist

### DIFF
--- a/docker/deployment.Dockerfile
+++ b/docker/deployment.Dockerfile
@@ -1,12 +1,12 @@
 FROM ubuntu:20.04
 
-RUN apt-get -y update && apt-get -y install ca-certificates curl
+RUN apt-get -y update && apt-get -y install ca-certificates netcat
 
 RUN mkdir -p /opt
 COPY typesense-server /opt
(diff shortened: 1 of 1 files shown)


honzasterba:
relies on 8108 being the port, this can be changed via command line args and env vars you either need to account for that in your code or this is just never going to get accepted

advdv:
@honzasterba I find the response discouraging for a contribution. The core of this PR is to have either netcat or curl installed in the container so users can easily setup their own health check. I think that should be acceptable.

iamriajul:
@honzasterba how about just having curl or `wget` is installed in the docker image, user will have the option to setup their own healthcheck based on their configurations/ports.

For example, I use healthcheck with every services I deploy with docker.
image
image
image
image
And more...

PR #1193, opened 2023-08-29
Synonym endpoint

krunal1313:
## Change Summary

- created new end-points for managing synonyms at top level, different from collection specific end-points
- making collection specific synonyms use top level synonym resource
- add synonym_sets in collection schema while creating collection which will be used to look for synonym while searching

## PR Checklist


kishorenc:
Will have to rethink this given the complication around migrating old end-point data etc. Will close this for now.

PR #1237, opened 2023-09-21
Add bulgarian language support

TihomirDobrev:


kishorenc:
We've already added this in the working branch for v0.26: 

https://github.com/typesense/typesense/blob/v0.26-facets/src/tokenizer.cpp#L340

PR #1356, opened 2023-11-03
Allow a field inside an object to have `reference`.

happy-san:
## Change Summary
Adds the ability to have a `reference` field inside an object. If we have an object `order` in the document and it has a reference to a `product`, the following schema can be specified:
```json
{
  "name": "Coll",
  "fields": [
    {"name": "order", "type": "object"},
    {"name": "order.product_id", "type": "string", "reference": "Products.id"}
  ]
}
```
If we had an array of objects instead with each object containing a reference, the type of the reference field would be array as well
```json
{
  "name": "Coll",
  "fields": [
    {"name": "order", "type": "object[]"},
    {"name": "order.product_id", "type": "string[]", "reference": "Products.id"}
  ]
}
```
The referenced document of an object is included inside the object itself instead of in the document.

## PR Checklist

### DIFF
--- a/src/collection.cpp
+++ b/src/collection.cpp
@@ -72,11 +72,36 @@ uint32_t Collection::get_next_seq_id() {
     return next_seq_id++;
 }
 
-Option<bool> Collection::add_reference_helper_fields(nlohmann::json& document) {
+Option<bool> single_value_filter_query(nlohmann::json& document, const std::string& field_name,
+                                       const std::string& ref_field_type, std::string& filter_query) {
+    auto const& value = document[field_name];
+    if (value.is_string() && ref_field_type == field_types::STRING) {
+        filter_query[filter_query.size() - 1] = '=';
+        filter_query += (" " + value.get<std::string>());
+    } else if (value.is_number_integer() && (ref_field_type == field_types::INT64 ||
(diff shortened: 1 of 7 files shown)


src/collection.cpp
```diff
@@ -72,11 +72,36 @@ uint32_t Collection::get_next_seq_id() {
     return next_seq_id++;
 }
 
-Option<bool> Collection::add_reference_helper_fields(nlohmann::json& document) {
+Option<bool> single_value_filter_query(nlohmann::json& document, const std::string& field_name,
+                                       const std::string& ref_field_type, std::string& filter_query) {
+    auto const& value = document[field_name];
```
kishorenc, line 77: Is `field_name` guaranteed to exist in document? (e.g. what if field is optional)

happy-san: Yes, `field_name` is guaranteed to exist. `single_value_filter_query` is only called in `add_reference_helper_fields` and we have these checks before calling `single_value_filter_query `.
* [Check 1](https://github.com/typesense/typesense/blob/de036ba406a2e2ceeeadf65b1c19438437132f58/src/collection.cpp#L105-L107)
* [Check 2](https://github.com/typesense/typesense/blob/de036ba406a2e2ceeeadf65b1c19438437132f58/src/collection.cpp#L217-L219)

kishorenc (approved):

Issue #1315, opened 2023-10-20
Out-of-the-box image embeddings


davidoort:
## Description

Would be great to be able to have the same workflow for image embeddings as for text embeddings. This would be very powerful for eCommerce use cases for instance where you want to show relevant products based on product and variant images as well as for "Similar to" searches where you can do image search and find similar items visually.

gustavopch:
I believe this can be closed.

Issue #1073, opened 2023-06-28
Feature request: Built-in image embeddings


gustavopch:
The idea is to enable searching images (or images + text). A good reference of how that could be done is https://marqo.ai. Instead of adding an image field, it could just read the URL from a string field and fetch the image server-side to generate the embedding.

kishorenc:
Reading from a remote URL is a bit wasteful and flaky. We could support upload and encoding of image into vectors but not store it. The image can be sent in base64 encoded form within JSON document. If we supported gzip encoding of import stream (work already in progress), then this should be efficient.

gustavopch:
And maybe resize the image down before uploading. Just guessing, but I suppose for most cases the quality of the search wouldn't change beyond a certain resolution. Maybe make the resolution configurable in the SDKs so people can send larger or smaller images according to their use cases.

gustavopch:
I believe this can be closed.

PR #1385, opened 2023-11-16
Add support for image embeddings using CLIP

ozanarmagan:
## PR Checklist


ozanarmagan:
#1073 & #1315

src/collection.cpp
```diff
@@ -624,7 +624,13 @@ void Collection::batch_index(std::vector<index_record>& index_records, std::vect
         if(index_record.indexed.ok()) {
             if(index_record.is_update) {
                 remove_flat_fields(index_record.new_doc);
+                for(auto& field: fields) {
```
kishorenc, line 627: Instead of looping here, let's see if we can do this processing inside `validate_index_in_memory` which is already looping on the schema fields.

kishorenc (changes requested):

kishorenc (approved):

PR #1505, opened 2024-01-23
add resource error response in /health end-point

krunal1313:
## Change Summary

- add `resource_error` response in /health end-point in case of out_of_memory or out_of_disk resource errors
## PR Checklist

### DIFF
--- a/include/cached_resource_stat.h
+++ b/include/cached_resource_stat.h
@@ -5,6 +5,13 @@
 #include <sys/statvfs.h>
 
 class cached_resource_stat_t {
+public:
+    enum resource_check_t {
+        OK,
+        OUT_OF_DISK,
+        OUT_OF_MEMORY
+    };
+
 private:
     const static size_t REFRESH_INTERVAL_SECS = 5;
     uint64_t disk_total_bytes = 0;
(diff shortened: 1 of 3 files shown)


include/cached_resource_stat.h
```diff
@@ -18,6 +18,11 @@ class cached_resource_stat_t {
 
     uint64_t last_checked_ts = 0;
 
+    static constexpr const char* out_of_memory_str = "OUT_OF_MEMORY";
+    static constexpr const char* out_of_disk_str = "OUT_OF_DISK";
+
+    std::string out_of_resource_error="";
```
kishorenc, line 24: This need not be a string, use `enum resource_check_t`

src/cached_resource_stat.cpp
```diff
@@ -89,8 +92,13 @@ cached_resource_stat_t::has_enough_resources(const std::string& data_dir_path,
...
                   << ", memory_free_min: " << memory_free_min_bytes;
+        out_of_resource_error = out_of_memory_str;
         return cached_resource_stat_t::OUT_OF_MEMORY;
     }
 
     return cached_resource_stat_t::OK;
 }
+
+const std::string cached_resource_stat_t::get_out_of_resource_error() const {
+    return out_of_resource_error;
```
kishorenc, line 103: We can use `magic_enum::enum_name(enum_value)` to convert the enum to string.

src/cached_resource_stat.cpp
```diff
@@ -89,8 +92,13 @@ cached_resource_stat_t::has_enough_resources(const std::string& data_dir_path,
         LOG(INFO) << "memory_total: " << memory_total_bytes << ", memory_available: " << memory_available_bytes
                   << ", all_memory_used: " << all_memory_used << ", free_mem: " << free_mem
                   << ", memory_free_min: " << memory_free_min_bytes;
+        out_of_resource_error = out_of_memory_str;
         return cached_resource_stat_t::OUT_OF_MEMORY;
     }
 
     return cached_resource_stat_t::OK;
```
kishorenc, line 99: We should also set to `OK` here.

kishorenc (changes requested):

Issue #1497, opened 2024-01-18
Feature Request: Configurable HNSW Parameters (M, efConstruction, efSearch) in Typesense for Improved Semantic Search


AmanChourasia-TomTom:
## Issue Description:
In my usage of Typesense's semantic search capabilities, I've observed that the default settings for the HNSW parameters (M, efConstruction, efSearch) are not producing optimal results. These defaults seem to be limiting the search accuracy during semantic search.

## Enhancement needed:
I request to enable user-configurable HNSW parameters (M, efConstruction, efSearch) in Typesense. This enhancement would offer greater flexibility in fine-tuning the search functionality, potentially improving the accuracy and response time of semantic search results.

DmitryKey:
Looking at the commit history -- this ticket has been addressed?

jasonbosco:
Yup. This is now available in `0.26.0.rc54`:

```json
{
  "name": "vector_search",
  "fields": [
    {
      "name": "vec",
      "type": "float[]",
      "num_dim": 768,
      "hnsw_params": {
        "ef_construction": 200,
        "M": 16
      }
    }
  ]
}
```

jasonbosco:
Released in v26

PR #1516, opened 2024-01-27
Make HNSW params configurable

ozanarmagan:
### Indexing Parameters
You should set `ef_construction` and `M` for vector and embedding fields while creating/altering the collection.

```json
{
  "name": "vector_search",
  "fields": [
    {
      "name": "vec",
      "type": "float[]",
      "num_dim": 768,
      "hnsw_params": {
        "ef_construction": 200,
        "M": 16
      }
    }
  ]
}
```
### Search Parameters
You can set `ef` parameter in vector query.
```
GET /collections/vector_search/documents/search?q=*&vector_query=vec:([], ef:100)
```


ozanarmagan:
#1497

PR #1545, opened 2024-02-12
return error response when stopword is not found

krunal1313:
## PR Checklist

### DIFF
--- a/src/stopwords_manager.cpp
+++ b/src/stopwords_manager.cpp
@@ -98,3 +98,9 @@ void StopwordsManager::dispose() {
     std::unique_lock lock(mutex);
     stopword_configs.clear();
 }
+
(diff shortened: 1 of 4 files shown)


src/stopwords_manager.cpp
```diff
@@ -97,4 +97,10 @@ Option<bool> StopwordsManager::delete_stopword(const std::string& stopword_name)
 void StopwordsManager::dispose() {
     std::unique_lock lock(mutex);
     stopword_configs.clear();
+}
+
+bool StopwordsManager::stopword_exists(const std::string &stopword) {
+    std::unique_lock lock(mutex);
```
kishorenc, line 103: Use shared lock

kishorenc (approved):

PR #1570, opened 2024-02-22
Fix sort fields bug for hybrid search with self-populated embeddings

ozanarmagan:
## PR Checklist

### DIFF
--- a/src/collection.cpp
+++ b/src/collection.cpp
@@ -1562,7 +1562,7 @@ Option<bool> Collection::validate_and_standardize_sort_fields(const std::vector<
         }
     }
 
-    if(!found_match_score && !is_wildcard_query && sort_fields.size() < 3) {
+    if(!found_match_score && !is_wildcard_query && sort_fields_std.size() < 3) {
         sort_fields_std.emplace_back(sort_field_const::text_match, sort_field_const::desc);
     }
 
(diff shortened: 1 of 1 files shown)


src/collection.cpp
```diff
@@ -1566,7 +1566,8 @@ Option<bool> Collection::validate_and_standardize_sort_fields(const std::vector<
         sort_fields_std.emplace_back(sort_field_const::text_match, sort_field_const::desc);
     }
 
-    if(!found_vector_distance && is_vector_query && sort_fields.size() < 3) {
+    // only add vector_distance if it is a semantic search, do not add it for hybrid search
+    if(!found_vector_distance && is_vector_query &% is_wildcard_query && sort_fields.size() < 3 && sort_fields_std.size() < 3) {
```
kishorenc, line 1570: Consider a case where there sort by is `is_promoted:true` and there a hybrid query is made using self-generated embedding and `vector_query`. We will end up adding both `text_match` and `vector_distance`. What we should instead do I think is this:

```
if(!found_match_score && !is_wildcard_query && sort_fields_std.size() < 3) {
    sort_fields_std.emplace_back(sort_field_const::text_match, sort_field_const::desc);
    found_match_score = true;
}

if(!found_vector_distance && !found_match_score && is_vector_query && sort_fields_std.size() < 3) {
    sort_fields_std.emplace_back(sort_field_const::vector_distance, sort_field_const::asc);
}
```

Let me know what you think.

kishorenc (approved):

Issue #780, opened 2022-11-09
[Feature Request] Ability to change maximum page size, beyond 250


shejal-Hauper-fe:
I need to add a feature that downloads the table record in CSV format when I click the export to csv button. I want to retrieve all of the data from the query, not just the first 250 entries. What are my options in this scenario?

This is the screenshot of the  error I'm facing:
![limitation-error](https://user-images.githubusercontent.com/103435307/200739567-31797161-d753-4eef-b18f-9022f88aac54.png)

This is the query string:
"searches": [
      {
        "query_by": "cryptoScoreCoarse,cryptoScore",
        "sort_by": "",
        "highlight_full_fields": "application",
        "collection": collection,
        "q": "*",
        "facet_by": "cryptoStrength,keyExchange,bulkEncryption,modeOfOperation,signature,hash,mac,protocol,zoneIn,zoneOut,clientos,serveros",
        "max_facet_values": 150,
        "page": 1,
        "per_page": 2000
        
      }
    ]

shejal-Hauper-fe:
I found one opened ticket regarding this:
https://github.com/typesense/typesense/issues/672
However, if we have 10M data and receive 250 each time, there would be 4000 requests (1,000,000/250), which could cause the system to crash.

jasonbosco:
You can use the documents export endpoint to export the entire dataset without pagination. But it only supports `filter_by` and not the `q` parameter.

shejal-Hauper-fe:
@jasonbosco thanks for the quick response. Now I want to search data and based on sorted result I want to export it in CSV what I can do for this?

mahmudtopu3:
It will be better if the limit is modifiable so that one can increase even though the performance will get down.

Clasyc:
Yes it would be great to have option on modifying the limit.

rahularoradfs:
+1

chrysos:
+1

garygreen:
## **Algolia** has a max limit of **1,000**

![image](https://github.com/typesense/typesense/assets/1702638/2dd2a55d-3973-4355-958d-36eefb70d8ad)

## **Meilisearch** has a max limit of **1,000**

![image](https://github.com/typesense/typesense/assets/1702638/ca1f9775-5de7-4499-afa6-18e799bdd925)

## **ElasticSearch** has a max limit of **10,000** (with a caution on memory usage, but leaves that in the users court)

[...]

PR #1577, opened 2024-02-25
Added config option max_per_page

lukaslau:
## Change Summary
Added config option max-per-page and default-topster-size to be able to increase max hits count per page.

Configs max-per-page and default-topster-size value should be the same.

## PR Checklist
https://github.com/typesense/typesense/issues/780


kishorenc:
Thank you for the PR. We are fine with making these configurable. However, we will have to run some benchmarks to understand whether making these parameters a variable ends up disturbing any compiler optimizations that are currently relying on them being static. 

We're in the last leg of the 0.26 release, so will test and benchmark this change in a few weeks.

kishorenc:
@lukaslau 

Happy to review this now that the v26 release is done. Can you please rebase this PR against the v27 branch? Thank you.

src/collection.cpp
```diff
@@ -2156,7 +2158,7 @@ Option<nlohmann::json> Collection::search(std::string raw_query,
         return Option<nlohmann::json>(422, message);
     }
 
-    size_t max_hits = DEFAULT_TOPSTER_SIZE;
+    size_t max_hits = Config::get_instance().get_default_topster_size();
```
kishorenc, line 2161: There is no need to parameterize `DEFAULT_TOPSTER_SIZE` because we assign max_hits in the next line this way:

```
max_hits = std::min(std::max(fetch_size, max_hits), get_num_documents());
```

Since `fetch_size` is already calculated based on the pagination depth, we will automatically have a larger max_hits value. 

So this PR only needs to parameterize `max_per_page`.

kishorenc (changes requested):

.github/workflows/tests.yml
```diff
@@ -99,4 +99,4 @@ jobs:
         uses: actions/upload-artifact@v3
         with:
           name: typesense-server
-          path: ${{ env.BAZEL_BIN_FULL_PATH }}/typesense-server
```
kishorenc, line 102: Unrelated change?

include/tsconfig.h
```diff
@@ -80,6 +80,8 @@ class Config {
 
     bool enable_search_logging;
 
+    int max_per_page;
```
kishorenc, line 83: Let's use `uint32_t` instead of `int` since the value cannot be negative.

kishorenc (changes requested):
Looks great, just a couple of minor changes and I can merge after that. The most important one is using `uint32_t` instead of `int` everywhere for max per page variable.

kishorenc (approved):
Thank you!

PR #1789, opened 2024-06-19
analytics db with ttl

krunal1313:
## Change Summary

- add ttl based analytics db to store log events
- add end-points to send events to all nodes
- add test

## PR Checklist


src/analytics_manager.cpp
```diff
@@ -742,4 +775,22 @@ void counter_event_t::serialize_as_docs(std::string &docs) {
     if (!docs.empty()) {
         docs.pop_back();
     }
+}
+
+bool AnalyticsManager::write_to_db(const nlohmann::json& payload) {
+    if(analytics_store) {
+        for(auto event: payload) {
+            std::string key = event["user_id"].get<std::string>() + "_" + std::to_string(event["timestamp"].get<uint64_t>());
```
kishorenc, line 783: Length of unix timestamps can increase over time. We should instead store them as uin64 bytes. Check `StringUtils::serialize_uint64_t`.

include/analytics_manager.h
```diff
@@ -53,13 +53,14 @@ struct event_t {
         }
     }
 
-    void to_json(nlohmann::json& obj) const {
+    void to_json(nlohmann::json& obj, const std::string& coll) const {
```
kishorenc, line 56: This function is now long enough that it's better off being in the `.cpp` file.

kishorenc (changes requested):

kishorenc (approved):

PR #1944, opened 2024-09-11
allow updating remote model api_key

krunal1313:
## Change Summary

- allow updating remote model api_key by patch endpoint of collection
- refactor collection metadata updation flow
- update related tests

### Updating api_key of a remote model
To update api_key of remote model, one need to send patch request to collection with model_name and new api_key.
Below is sample curl request to update api_key of companies collection.

```curl
curl "http://localhost:8108/collections/companies" \
       -X PATCH \
       -d '{
         "fields": [
          {
            "name": "embedding",
            "embed": {
              "from": ["product_name"],
              "model_config": {
                "model_name": "openai/text-embedding-3-small",
                "api_key": "haskhcxkkck21989878dsd-asdkjasjcbb"
              }
            }
          }
        ]
       }'
```
Here `api_key` is the new api_key to be updated.
All fields are mandatory; the update request will not succeed without them.

## PR Checklist

### DIFF
--- a/src/collection_manager.cpp
+++ b/src/collection_manager.cpp
@@ -2291,18 +2291,3 @@ Option<bool> CollectionManager::update_collection_metadata(const std::string& co
 
     return Option<bool>(400, "failed to insert into store.");
 }
-
-Option<bool> CollectionManager::update_model_apikey(const std::string& collection, const nlohmann::json& model_config) {
-    auto collection_ptr = get_collection(collection);
-    if (collection_ptr == nullptr) {
-        return Option<bool>(400, "failed to get collection.");
-    }
-
-    auto op = collection_ptr->update_apikey(model_config);
-
-    if (!op.ok()) {
-        return op;
-    }
-
-    return Option<bool>(true);
-}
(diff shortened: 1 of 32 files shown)


src/collection_manager.cpp
```diff
@@ -2269,7 +2269,11 @@ bool CollectionManager::is_valid_api_key_collection(const std::vector<std::strin
 }
 
 bool CollectionManager::update_collection_metadata(const std::string& collection, const nlohmann::json& metadata) {
+    auto collection_ptr = get_collection(collection);
+    collection_ptr->update_metadata(metadata);
```
kishorenc, line 2273: Need to check if `collection_ptr` is null here.

src/collection_manager.cpp
```diff
@@ -2278,4 +2282,14 @@ bool CollectionManager::update_collection_metadata(const std::string& collection
     collection_meta_json[Collection::COLLECTION_METADATA] = metadata;
 
     return store->insert(collection_metakey, collection_meta_json.dump());
+}
+
+Option<bool> CollectionManager::update_model_apikey(const std::string& collection, const nlohmann::json& model_config) {
+    auto op = get_collection(collection)->update_apikey(model_config);
```
kishorenc, line 2288: Check if get_collection returns non-null.

kishorenc (changes requested):

kishorenc (approved):

PR #2127, opened 2025-01-02
truncate all docs in collection

krunal1313:
## Change Summary

- add end-point `/collections/:collection/documents/truncate_all` to remove all docs from collection
- add test

### Truncate all docs from collection
To truncate all docs from a collection, one need to put a DELETE request to end-point `/collections/:collection/documents/truncate_all` like below,

```curl
curl -H "X-TYPESENSE-API-KEY: ${TYPESENSE_API_KEY}" -X DELETE  "http://localhost:8108/collections/hnstories/documents/truncate_all"
```
Here, `hnstories` is the collection to truncate all docs.

And we get the response like below,

```json
{"num_deleted":1000000}
```
Here `num_deleted` is the count of docs deleted.

In case of empty collection we get response accordingly,
```json
{"num_deleted":0}
```

## PR Checklist

### DIFF
--- a/src/collection.cpp
+++ b/src/collection.cpp
@@ -7505,6 +7505,48 @@ Option<nlohmann::json> Collection::get_alter_schema_status() const {
     return Option<nlohmann::json>(status_json);
 }
 
+Option<size_t> Collection::remove_all_docs() {
+    size_t num_docs_removed = 0;
+
+    const std::string seq_id_prefix = get_seq_id_collection_prefix();
+    std::string upper_bound_key = get_seq_id_collection_prefix() + "`";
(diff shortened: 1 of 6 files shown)


src/collection.cpp
```diff
@@ -7505,6 +7505,53 @@ Option<nlohmann::json> Collection::get_alter_schema_status() const {
...
+                LOG(INFO) << "Removed " << num_docs_removed << " so far.";
+            }
+        }
+
+        iter->Next();
+    }
+
+    if(num_docs_removed) {
+        store->flush();
+        store->compact_all();
```
kishorenc, line 7549: We should do `store->compact_range(del_key_prefix, del_end_prefix);`

PR #2204, opened 2025-02-17
Query Analytics with meta fields

krunal1313:
## PR Checklist

## Query Analytics with meta fields
Typesense supports query analytics with meta fields. Currently limited to `filter_by` and `analytics_tags`.
Here, `analytics_tags` are tags passed with search query for analytics purpose.
### Illustration
First create a collection to search on,
```json
{
            "name": "hnstories",
            "fields": [
                {"name": "points", "type": "int32"},
            ]
 }
```
Add some documents to the collection.
Now create suggestion collection to store queries. For storing `filter_by` string and `analytics_tags`, we need to add that in the schema as below,
```json
{
        "name": "top_queries",
        "fields": [
          {"name": "q", "type": "string" },
          {"name": "filter_by", "type" : "string"},
          {"name": "analytics_tags", "type" : "string[]"},
          {"name": "count", "type": "int32" }
        ]
 }
```
Let's add analytics rule to store popular queries in above collection. To enable meta fields analytics, we need to pass array of supported meta fields like below,
```json
{
        "name": "top_search_queries",
        "type": "popular_queries",
        "params": {
            "limit": 100,
            "expand_query": true,
            "meta_fields":["analytics_tags", "filter_by"],
            "source": {
                "collections": ["titles"]
            },
            "destination": {
                "collection": "top_queries"
            }
        }
 }
```

Now when we search on the collection with `filter_by` string and `analytics_tags` like below,
```curl
curl -H "X-TYPESENSE-API-KEY: ${TYPESENSE_API_KEY}" "http://localhost:8108/collections/hnstories/documents/search?q=iron&query_by=title&filter_by=points:>100&analytics_tags=oxford"
```
This will aggregate the queries with count and will save `filter_by` string and `analytics_tags` also.
Upon checking the docs in `top_queries` collection, we get below response,

```json
{"facet_counts":[],"found":1,"hits":[{"document":{"count":1,"analytics_tags":["london"], "filter_by":"points:>100","id":"9677940243608161791","q":"iron"},"highlight":{},"highlights":[]}],"out_of":1,"page":1,"request_params":{"collection_name":"product_queries","first_q":"*","per_page":10,"q":"*"},"search_cutoff":false,"search_time_ms":0}
```

##### Notes:
- to use meta fields analytics, destination collection should have respective meta fields in schema viz. `filter_by` and `analytics_tags` and analytics rule should have array of meta fields
- failing above will result in not storing meta fields


src/query_analytics.cpp
```diff
@@ -57,11 +57,39 @@ void QueryAnalytics::serialize_as_docs(std::string& docs) {
...
     std::string key_buffer;
+    nlohmann::json doc;
     for(auto it = local_counts.begin(); it != local_counts.end(); ++it) {
+        std::vector<std::string> results;
+
         it.key(key_buffer);
-        nlohmann::json doc;
+        //sample string : Nike+filter=Air+tags=Colaba,Bandra
+        StringUtils::split(key_buffer, results, "+");
+        if(meta_fields_analytics) {
```
kishorenc, line 67: This is a bit complex and is necessitated only because the `local_counts` map is using `std::string` as the key. We should instead have a struct as the key of the local_counts map. The struct could look like:

```
struct analytics_params_t {
  std::string query;
  std::map<std::string, std::string> meta;
}
```
We can then have a custom hash function for the map. 

Both `filter_by` and `tag` can go into the meta map. Also, we decided that a single tag string is enough: no need to support array of tags. So let's call it as `analytics_tag`.

include/query_analytics.h
```diff
@@ -9,11 +9,12 @@
 
 class QueryAnalytics {
 public:
-    struct QWithTimestamp {
+    struct QWithTimestampFilter {
```
kishorenc, line 12: We can undo this change.

include/collection.h
```diff
@@ -918,7 +922,8 @@ class Collection {
                                   const size_t& max_filter_by_candidates = DEFAULT_FILTER_BY_CANDIDATES,
                                   bool rerank_hybrid_matches = false,
                                   bool validate_field_names = true,
-                                  bool enable_analytics = true) const;
+                                  bool enable_analytics = true,
+                                  std::string analytics_tags="") const;
```
kishorenc, line 926: Need to support only a single tag, so change to `analytics_tag` (see explanation in the other review comment)

kishorenc (changes requested):

kishorenc (approved):

PR #2216, opened 2025-02-22
Fix typesense-server.service

MrShitFox:
When installing the deb package I noticed that the path in the service is incorrect, this is a fix.

### DIFF
--- a/debian-pkg/typesense-server/etc/systemd/system/typesense-server.service
+++ b/debian-pkg/typesense-server/etc/systemd/system/typesense-server.service
@@ -5,7 +5,7 @@ Requires=network.target remote-fs.target
 After=network.target remote-fs.target
 
 [Service]
-ExecStart=/usr/bin/typesense-server --config=/etc/typesense/typesense-server.ini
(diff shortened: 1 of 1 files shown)


kishorenc:
We do actually have `typesense-server.ini` here: https://github.com/typesense/typesense/blob/v29/debian-pkg/typesense-server/etc/typesense/typesense-server.ini

We have not modified this in some time: can you please tell me where you are installing the DEB from?

MrShitFox:
I installed the deb package from the link that is currently listed on the official site, namely “https://dl.typesense.org/releases/28.0/typesense-server-28.0-amd64.deb”. After installation I could not understand why the configuration was not applied, then I found out that the config “typesense.ini” did not match the one specified in the systemd service. I describe what I had when installing this particular version.

kishorenc:
Can you try this again on a fresh install? We have had a lot of installs of this DEB, including from within Typesense Cloud and the path is correct (see the Github link in my previous link).

kishorenc:
Going to close this since I don't think the service file name is actually wrong. Feel free to reopen if you can show some reproducible steps for it -- maybe there is an edge case that I am not accounting for.

Issue #828, opened 2022-12-19
Filter for two properties within a nested object / field


kishorenc:
## Description

Creating a separate issue for the feature ask on this [comment](https://github.com/typesense/typesense/issues/227#issuecomment-1211828348).

-----

I'd like to filter for two properties within an object that is part of an array of objects.

For example:

Document:

```
{
    "name": "Chili con carne",
    "ingredients": [{
       "name": "Ground Beef",
       "concentration": 30
    }]
}
```

I'd like to search for this meal like this:

```
search_parameters = {
    'q':              '*',
    'query_by':       'ingredients',
    'filter_by':      'ingredients.name: Ground Beef && ingredients.concentration:>= 20',
    'include_fields': 'ingredients'
}
```

So I'd like to filter for an ingredient that has a concentration of "Ground beef" of at least 20%.

Currently, the query above seems to look for all concentration estimates within the ingredients array and not just the one within the "ground beef" object.

For example, it will match this document, when it should **not**:

```
{
    "name": "Tomato Soup",
    "ingredients": [
      {
        "name": "Ground Beef",
        "concentration": 10
      },
      {
        "name": "Tomato",
        "concentration": 30
      },
    ]
}
```

-----

This might require a new filter by syntax where we can scope a bunch of filter expressions into a given object's sub fields.

DannyCarvajal:
I would really appreciate having this feature for our project! awesome :)

JasonAlvarado:
This feature can be a good  solution to simulate a search in 2 collections

jperasmus:
Adding our interest for this feature as well.

I see ElasticSearch solves this problem with a special "nested" object type that under the hood creates separate indexes for each object in the array - essentially treating them as sub-documents that can be matched against as a whole.

https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html

scott-wiley:
Very interested in this feature!

jsutaria:
+1

PlkMarudny:
I think the problem is even bigger. I have a document, that has some metadata attached in a form array of labels with confidence assigned:

```
[ { label: Red, confidence: 75.243}, ...]
```

Now, when I issue a query like:
```
filter_by: persons._id:=[60607ae22223826df51be70b] && labels.label:=Red  && labels.confidence:>85 && labels.label:=Blazer
```
I receive messed up results, where ```labels.confidence``` is ignored, but the following query:
```
[...]

PR #2268, opened 2025-03-21
filtering on fields inside nested object/field

krunal1313:
## PR Checklist

## Filtering on fields inside a nested field/object
Syntax : `<nested_field_parent>.{<nested_field1> : <val> ...}`

Let's suppose we've the following schema,
```json
{
      "name": "menu",
      "fields": [
           {"name": "name", "type": "string"},
           {"name": "ingredients", "type": "object[]"},
           {"name": "ingredients.*", "type": "auto", "optional": true}
       ],
       "enable_nested_fields": true
}
```
And following records,
```json
{ 
   "name": "Pasta",
   "ingredients": [{"name": "cheese", "concentration": 40}, {"name" : "spinach", "concentration": 10}, {"name": "jalepeno", "concentration": 20}]
}
{
   "name": "Pizza",
   "ingredients": [{"name": "cheese", "concentration": 30}, {"name": "pizza sauce", "concentration": 30}, {"name": "olives", "concentration": 30}]
}
{
   "name": "Lasagna",
   "ingredients": [{"name": "cheese", "concentration": 60}, {"name": "jalepeno", "concentration": 20}, {"name": "olives", "concentration": 20}]
}
```

Now to filter using nested fields of `ingredients` we've to use special filter_by query syntax like below,
`filter_by=ingredients.{name : cheese && concentration :<50}`
Above filter_by string will filter results with records `Pizza` and `Pasta`

Likewise, 
we can pass all kind of filter_by string expressions inside `{}`
filter_query can be extended to multiple nested object filters too.
#### Note : should be mindful on syntax part for targeting such filtering. Going by traditional filter_by syntax `ingredients.name: cheese && ingredients.concentration:<50` will match all 3 records.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -255,8 +255,8 @@ new_git_repository(
     name = "icu",
     build_file = "//bazel/icu:BUILD",
     patches = ["//bazel/icu:icu.patch"],
-    remote = "https://github.com/unicode-org/icu.git",
-    tag = "release-71-1",
+    remote = "https://github.com/typesense/icu.git",
+    tag = "release-71-1-patch",
 )
 
 git_repository(
--- a/include/filter.h
+++ b/include/filter.h
@@ -6,7 +6,12 @@
 #include "json.hpp"
 #include "store.h"
 
-constexpr uint32_t COMPUTE_FILTER_ITERATOR_THRESHOLD = 25'000;
+#ifdef TEST_BUILD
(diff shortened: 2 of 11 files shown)


kishorenc:
Fixes https://github.com/typesense/typesense/issues/828

kishorenc:
We found an issue that required a revert, but we have now identified the root cause. We are now addressing it, and will be releasing a new RC build with the feature intact.

Issue #2172, opened 2025-02-05
The `queries` configuration option in the `vector_query` parameter doesn’t adhere to any special character escape strategy.


AdamAld:
# Description
Adding a string containing special characters to the `queries` option of `vector_queries` is being parsed as two different entries. Using backticks - as referenced in the [Tips for Filtering](https://typesense.org/docs/guide/tips-for-filtering.html#escaping-special-characters) guide - is not respected.

# Steps to reproduce
```shell
### Run Typesense via Docker ########################################
export TYPESENSE_API_KEY=xyz

mkdir "$(pwd)"/typesense-data

docker run -p 8108:8108 \
            -v"$(pwd)"/typesense-data:/data typesense/typesense:27.1 \
            --data-dir /data \
            --api-key=$TYPESENSE_API_KEY \
            --enable-cors

### Reproduction Steps ###############################################
export TYPESENSE_API_KEY=xyz

curl "http://localhost:8108/debug" \
       -H "X-TYPESENSE-API-KEY: ${TYPESENSE_API_KEY}"

curl "http://localhost:8108/collections" \
       -X POST \
       -H "Content-Type: application/json" \
       -H "X-TYPESENSE-API-KEY: ${TYPESENSE_API_KEY}" \
       -d '{
         "name": "companies",
         "fields": [
           {"name": "id", "type": "string" },
           {"name": "someObj.summary", "type": "string[]", "optional": false, "index": true },
           {"name": "someObjEmbeddings", "type":"float[]", "embed": { "from": ["someObj.summary"], "model_config": { "model_name": "ts/all-MiniLM-L12-v2" } } },
           {"name": "updatedAt", "type": "int64" }
         ],
         "enable_nested_fields": true,
         "token_separators": ["(", ")", "-", "."],
         "default_sorting_field": "updatedAt"
       }'

curl "http://localhost:8108/collections/companies/documents/import?action=upsert" \
        -H "X-TYPESENSE-API-KEY: ${TYPESENSE_API_KEY}" \
        -H "Content-Type: text/plain" \
        -X POST \
        -d '{ "id": "1", "someObj.summary": ["Something something 1,4-Butanediol with some other stuff"], "updatedAt": 1735837189107 }
            { "id": "2", "someObj.summary": ["Different info about jazz not related to anything else"], "updatedAt": 1733782838569 }'

curl "http://localhost:8108/multi_search" \
        -X POST \
        -H "Content-Type: application/json" \
[...]

PR #2273, opened 2025-03-24
Fix vector query parsing with escape

ozanarmagan:
## Usage

```
vector_field:([], queries:[query_one, `this, query, will be, a single, query`)
```


ozanarmagan:
Fix #2172

kishorenc (approved):

PR #2320, opened 2025-04-28
Batch Single doc write requests

krunal1313:
## PR Checklist

### DIFF
--- a/include/collection.h
+++ b/include/collection.h
@@ -451,6 +451,14 @@ private:
 
     std::deque<nlohmann::json> alter_history;
 
+    const int batch_interval = -1;
+
+    std::vector<std::string> single_doc_batch;
+
+    std::chrono::steady_clock::time_point last_batch_flush_secs;
+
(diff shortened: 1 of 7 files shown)


src/async_doc_request.cpp
```diff
@@ -0,0 +1,144 @@
...
+                auto collection = CollectionManager::get_instance().get_collection(coll_name);
+                auto async_reqs = kv.second;
+                nlohmann::json document;
+
+                std::vector <std::string> json_lines;
+                for (const auto& async_req: async_reqs) {
+                    json_lines.push_back(async_req.first);
+                }
+
+                nlohmann::json json_res = collection->add_many(json_lines, document, index_operation_t::CREATE, "",
```
kishorenc, line 29: We can't always assume that all the writes will be of type `index_operation_t::CREATE`. We will have to see how to handle this. Definitely the stream of async writes can involve a mix of create, update and upsert.

kishorenc (changes requested):

Issue #2205, opened 2025-02-17
"filter_by" query returning non-matching results


jeanmatthieud:
I'm doing a basic request on a field:
```json
{
  "q": "*",
  "filter_by": "option_section:=`140 x 140 mm`",
  "page": 1,
  "per_page": 10,
  "exhaustive_search": false
}
```

I also tried:
```json
{
  "q": "*",
  "filter_by": "option_section:=140 x 140 mm",
  "page": 1,
  "per_page": 10,
  "exhaustive_search": false
}
```

But the results don't match the request (see file attached with the first page of results).
It does work for some other section values.

[export.json](https://github.com/user-attachments/files/18825254/export.json)

# Typesense version
- Tried on v28.0 and v27.1
- Tried on a freshly new created collection on v28.0 and an old on with v27.1 and v28.0
- Collection (the name differs from the results but it is the same collection that I recreated) : [product.json](https://github.com/user-attachments/files/18825283/product.json)

jeanmatthieud:
Maybe linked to https://github.com/typesense/typesense/issues/1226 ?

jeanmatthieud:
I also tried on v0.25.2, same issue

Issue #2198, opened 2025-02-13
Feature Request - Support for Multi-Word Ordered Filtering in filter_by


rajaie:
# Support for Multi-Word Ordered Filtering in `filter_by`

## Description

Currently, Typesense does not support filtering a string field on multiple space-separated words while ensuring that the words appear in the exact order within the field. The `:=` operator enforces an exact match, but it does not allow for partial matches where the specified phrase is contained within a longer string.

For example, using `filter_by: title:=hello world` would only return exact matches for "hello world" but would exclude documents where "hello world" appears as part of a longer title, such as:

- "introduction hello world welcome"
- "welcome to hello world"

A desired behavior would be to allow filtering for phrases appearing in order while still permitting additional text before or after, similar to how `q="hello world"` works but within `filter_by`.

## Proposed Enhancement

1. Extend `filter_by` functionality to allow multi-word ordered filtering within a string field.
2. Enable support for partial matches where the specified phrase appears in order anywhere in the string.
3. Consider adding an option for both prefix and postfix filtering, such as:
   ```typescript
   filter_by: title:=*hello world*
   ```
   to match titles that contain "hello world" in order, regardless of surrounding text.

## Current Workarounds & Limitations

- The `:=` operator can enforce exact matches but does not allow partial matches.
- Prefix filtering (`:=hello world*`) only solves half the issue and does not allow matches where "hello world" appears in the middle or end.
- Multi-search can be used with queries like `[{q: '"hello world"'}, {q: '"foo bar"'}]`, but it requires client-side handling of pagination and offsets.
- Multi-search currently functions as an OR operation, making it difficult to apply additional AND conditions across multiple fields.

## Use Case

A common use case is job search filtering, where users may want to find job titles or descriptions containing specific ordered phrases. For example:

```typescript
(title:`hello world` || title:`foo bar`) && (description:`hello world` || description:`foo bar`)
```

would require filtering across multiple fields while maintaining phrase order constraints.

## References

- Related feature request: [Issue #1219](https://github.com/typesense/typesense/issues/1219)
- [Union Search Documentation](https://github.com/typesense/typesense-website/blob/v28.0/docs-site/content/28.0/api/federated-multi-search.md#union-search)

Would love to hear your thoughts on whether this is feasible and if there are any potential workarounds in the meantime!

Issue #1219, opened 2023-09-13
Feature request: multi word "contains" on filter_by


jsutaria:
I have an array of job titles in my documents, and I'm looking to try to see if any of the job titles contains a substring of words. 

I.e. the query I'm making is `work_experiences.title: engineering manager`
I'd love to be able to match on titles in the filter_by for "software engineering manager" or "engineering manager team lead", and not match on a document that has ["engineering support team", "music manager"] as the array of titles.

shukryzablah:
+ 1 I think just ran into the same problem. It would be great to be able to filter `title:=[`engineering manager`, `founder`]` and it match on "engineering manager" OR founder exactly, not just parts of the literal

PR #2464, opened 2025-07-31
feat(filter): support phrase matching in filter_by 

felipehertzer:
## Change Summary

Hi, I'm having an issue filtering documents when using a compound word such as **"Formula One"**.

Currently, filtering behaves as follows:

* **Query:** `filter_by: text: Formula One`
* **Result:** Returns documents that contain both words separately, for example:

> "The scientist created a new formula, and this was just one of many groundbreaking discoveries in the lab."

What I expected was to only retrieve documents that contain the exact phrase:

> "Lewis Hamilton has won multiple Formula One World Championships."

I noticed that Typesense already supports the use of backticks to escape characters, so as an initial attempt, I reused the same syntax:

```text
filter_by: 'text: `Formula One`'
```

Close:
#2205 #2198 #1219

Thanks.

## PR Checklist

### DIFF
--- a/src/collection.cpp
+++ b/src/collection.cpp
@@ -5036,7 +5036,10 @@ void Collection::highlight_result(const bool& enable_nested_fields, const std::v
             text = document[search_field.name][match_index.index];
         }
 
-        handle_highlight_text(text, normalise, search_field, false, symbols_to_index, token_separators,
+        const auto& field_symbols = search_field.symbols_to_index.empty() ? symbols_to_index : search_field.symbols_to_index;
+        const auto& field_separators = search_field.token_separators.empty() ? token_separators : search_field.token_separators;
+
+        handle_highlight_text(text, normalise, search_field, false, field_symbols, field_separators,
                               highlight, string_utils, use_word_tokenizer, highlight_affix_num_tokens,
                               qtoken_leaves, last_valid_offset_index, prefix_token_num_chars,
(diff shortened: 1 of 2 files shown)


kishorenc:
Thank you, this will be a useful feature. We will be reviewing the PR next week.

felipehertzer:
Hey @kishorenc,

I’ve fixed the test errors and added additional test cases as well. There is still one failing test on my machine, which I believe may be related to a pointer or an environment-specific issue in `CollectionSortingTest.TestVectorQueryDistanceThresholdSorting`.

```
Expected equality of these values:
  0.07853113859891891
    Which is: 0.0785311
  res["hits"][0]["vector_distance"].get<float>()
    Which is: 0.0785313
```

happy-san:
Also, we have to support lazy evaluation of the filter_by expression. You can refer [this test](https://github.com/typesense/typesense/blob/v30/test/filter_test.cpp#L486-L502) for a better understanding. 
The data in the test won't induce lazy evaluation behavior. Maybe we should have documents having following values:
```
foo
foo bar baz
foo bar
bar
foo bar baz
baz
foo baz bar
foo bar baz
```
and the query:
```
q: foo
filter_by: field: "foo bar baz"
enable_lazy_filter: true
```
which should match 3 documents. Let me know if something doesn't make sense.

Issue #2238, opened 2025-03-04
Cluster stuck with no leader after DR changes all node IPs (Raft references old addresses)


jchaconst1:
## Description

During our **Disaster Recovery (DR)** process, **all** Typesense nodes change their IP addresses (e.g., from `x.x.26.x` to `x.x.27.x`). Once in the DR environment, we stop the Typesense service on each node and modify only the `.ini` file (specifically the `nodes` setting) to reference hostnames. However, Raft persists the old IPs in its logs/snapshots, causing the cluster to remain in a no-leader state after restart.

---

### Steps to Reproduce

1. **Set up a multi-node cluster** (e.g., 3 nodes in `x.x.26.x`) in the primary data center, each referencing the others in the `nodes` config.  
2. **Trigger a DR failover** so that all nodes come online in `x.x.27.x`.  
3. **Stop the service** on each node and edit the `.ini` to reference the new hostnames (or IPs).  
4. **Restart the service** with the updated `.ini`. Despite this, the Raft snapshots/logs still reference the old IPs, leaving the cluster without a leader.

---

### Expected Behavior

- That updating `nodes` in the `.ini` and restarting would allow the cluster to seamlessly transition to the new IP/hostname environment.  
- A supported method for Raft to remove or replace the old IP references when all nodes have changed addresses.

---

### Actual Behavior

- Even after editing `.ini`, Raft keeps the old `x.x.26.x` IP addresses in its internal state.  
- Leader elections fail, and the cluster remains stuck with messages like:
  ```
  WYYYYMMDD HH:MM:SS raft_server.cpp:734] Multi-node with no leader: refusing to reset peers.
  WYYYYMMDD HH:MM:SS node.cpp:1589] node default_group:x.x.27.x:8107:443 can't do pre_vote
    as it is not in x.x.26.11:8107:443, x.x.26.12:8107:443, x.x.26.13:8107:443
  ```

---

### Troubleshooting Attempts

1. **Removing/Editing Raft files** (e.g., `raft_meta`, `log_inprogress_...`):  
   - Often leads to data inconsistencies or collection loss because the Raft state no longer matches the on-disk data.  
2. **Switching to hostnames in `.ini`**:  
   - Does not help if the Raft log was already populated with old IP references, as the log seems to override the new configs.  
3. **Partial re-bootstrapping**:  
   - Requires carefully stopping and restarting nodes in a new configuration, which risks data integrity if not done meticulously.

---

### Current Workaround

[...]

PR #2496, opened 2025-08-15
[🛶 Raft Refactor Phase 1] Split raft_server.cpp code into modules

johnnyshields:
My team at TableCheck is trying to deploy Typesense at scale. We love the core featureset of TS but we have found the HA support is very poor and are only able to really run it as a Hot/Warm confirm (1 node actually serving requests.)

I'm going to do a series of PRs to improve the raft functionality that should be individually reviewable and mergeable,

In this PR, I've split `src/raft_server.cpp` into 5 modules for better maintainability and separation of concerns:

- `raft_server.cpp` - Slim coordinator
- `raft_config_manager.cpp` - DNS resolution and configuration parsing
- `raft_http_handler.cpp` - HTTP request processing and gzip handling
- `raft_lifecycle_manager.cpp` - Snapshot management and database lifecycle
- `raft_node_manager.cpp` - Node status monitoring and peer management

Here I've preserved all existing functionality and method names/signatures, just moved stuff around. No new tests added here since no new methods were added.

Next steps:
- [ ] Convert these modules to classes + tests for each class
- [ ] Add dynamic DNS resolution support (currently DNS is resolved once when node is added, but IPs for hostnames can change as per https://github.com/typesense/typesense/issues/2238)
- [ ] Immediate failure recovery (immediately trigger node refresh if change detected)
- [ ] Improve quorum/leader election, using monotonically versioned node config snapshots
- [ ] Add TLA+ spec for formal verification of quorum stuff


kishorenc:
Hi @johnnyshields 

> We love the core featureset of TS but we have found the HA support is very poor and are only able to really run it as a Hot/Warm confirm (1 node actually serving requests.)

Are you using Kubernetes? The challenges are mostly related to running on Kubernetes where pods can get replaced dynamically, with their IPs changing. The raft library we rely on (`braft`) still works with raw IP address, which are used as node identifiers. With a carefully implemented operator, Typesense can run well on Kubernetes. I recommend using: https://github.com/akyriako/typesense-operator

PR #2505, opened 2025-08-18
Fix dynamic filter override matching

krunal1313:
## Change Summary

- fix dynamic filter override matching
- add tests
- refactor sort_by during dynamic overrides matching
- fix when matched override is not having filter_by but search query has one

## PR Checklist


alangmartini:
Fixes https://github.com/typesense/typesense/issues/2506
Available in v30.0.rc9

Issue #2173, opened 2025-02-05
Ability to filter out duplicates with union


youradds:
I'm using the `union` feature in `multi_search` so that we can effectively do an "OR" search with multiple keywords/ phrases in one go. That works well - but the issue I'm having is that it duplicates a lot of stuff. Lets say I have 2 records:
```

ID: 1
title: test computer man

ID: 2
title computer geek
```

When I do a mutli_search for "**geek**" and "**computer**" and "**man**" (so 3 searches), I get 4 results - 2 for "computer" and 1 for "geek". and 1 for "man". So what I'm proposing is the ability to de-dupe them. Currently, I'm doing it in my perl code with :

```
        foreach (@{$results->{hits}}) {
            next if $seen_links->{$_->{document}->{id}};

            $seen_links->{$_->{document}->{id}} = 1;

            push @link_ids, $_->{document}->{id};

        }
```

But that obviously doesn't work super well - as it messes up the "per page" limit due to removing some of the records

Hopefully that makes sense!

aditio-eka:
Nice to see this PR! I’ve faced the same problem. Can I use this feature in `30.0.rc11`?

alangmartini:
Available in v30

PR #2511, opened 2025-08-20
Filter out duplicates with Union

krunal1313:
### Filter out duplicates with Union
When using `union` with multi-search, there might some instances when multiple keywords matched through different searches for same doc and those are duplicated in final response.
With optional flag `remove_duplicates`, one can enable/disable removal of these duplicated results.

#### By default this flag is set to `true` if not specified and has to be explicitly disabled by setting to `false` if required.

sample query :
```curl
curl 'http://localhost:8108/multi_search?page=1&per_page=2' -X POST \
     -H "X-TYPESENSE-API-KEY: ${TYPESENSE_API_KEY}" -d '
{
  "union": true,
  "remove_duplicates": <true/false>,
  "searches": [
    {
      "collection": "posts",
      "q": "comment",
      "query_by": "title"
    },
    {
      "collection": "posts",
      "q": "post",
      "query_by": "title"
    }
  ]
}'
```
## PR Checklist

### DIFF
--- a/include/collection.h
+++ b/include/collection.h
@@ -969,7 +969,7 @@ public:
 
     static Option<bool> do_union(const std::vector<uint32_t>& collection_ids,
                                  std::vector<collection_search_args_t>& searches, std::vector<long>& searchTimeMillis,
-                                 const union_global_params_t& union_params, nlohmann::json& result);
+                                 const union_global_params_t& union_params, nlohmann::json& result, bool remove_duplicates);
 
     Option<bool> get_filter_ids(const std::string & filter_query, filter_result_t& filter_result,
                                 const bool& should_timeout = true, const bool& validate_field_names = true) const;
--- a/include/collection_manager.h
+++ b/include/collection_manager.h
@@ -162,7 +162,7 @@ public:
 
     static Option<bool> do_union(std::map<std::string, std::string>& req_params,
                                  std::vector<nlohmann::json>& embedded_params_vec, nlohmann::json searches,
-                                 nlohmann::json& response, uint64_t start_ts, bool remove_duplicates = false);
+                                 nlohmann::json& response, uint64_t start_ts, bool remove_duplicates = true);
 
     static bool parse_sort_by_str(std::string sort_by_str, std::vector<sort_by>& sort_fields);
 
--- a/include/collection_manager.h
+++ b/include/collection_manager.h
@@ -162,7 +162,7 @@ public:
 
     static Option<bool> do_union(std::map<std::string, std::string>& req_params,
                                  std::vector<nlohmann::json>& embedded_params_vec, nlohmann::json searches,
-                                 nlohmann::json& response, uint64_t start_ts);
+                                 nlohmann::json& response, uint64_t start_ts, bool remove_duplicates = false);
 
     static bool parse_sort_by_str(std::string sort_by_str, std::vector<sort_by>& sort_fields);
 
--- a/include/topster.h
+++ b/include/topster.h
@@ -169,9 +169,11 @@ struct KV {
 
 struct Union_KV : public KV {
     uint32_t search_index{};
+    uint32_t collection_id;
+    bool remove_duplicates = false;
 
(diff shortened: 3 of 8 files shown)


happy-san:
#2173

kishorenc (approved):

kishorenc:
Sorry about the confusion here @aditio-eka -- while this PR was merged some time ago, we have not yet published it as part of the v30 RC builds so far. Recent RCs only carry some selected bug fixes, but this is not part of it. 

I will update this thread with a new build.

youradds:
> I've published `30.0.rc13` which contains the fix @aditio-eka .

Where do I get the v 30.0.rc13 from? I'm coming back to a site where I have Typesense, and this was one of those things that annoyed me :) (for example, I have a query with 13000 results, set for 30 per page - yet only 10 show, as they have had to be dedupted so we don't have them showing more than once)


UPDATE: I just realised how bad this is on some terms. For example, we have 13,051 results for "Halloween". The first page, returns just 12 - but it should be 55.  Then, based on that - 55 per page, we should have 236 pages, yet it runs out on page 76 :/ I didn't realise how bad it was on those terms :( 

Thanks

youradds:
> I've published `30.0.rc13` which contains the fix @aditio-eka .

I managed to get it going with the docker. So on v30, I pass:

`$VAR1 = {
          'union' => \1,
          'searches' => [
                          {
                            'q' => 'haunted'
                          },
                          {
                            'q' => 'ghost'
                          },
                          {
                            'q' => 'ghosts'
                          },
                          {
                            'q' => 'satan'
                          },
                          {
                            'q' => 'devil'
                          },
                          {
                            'q' => 'devils'
                          },
                          {
                            'q' => 'evil'
                          },
                          {
                            'q' => 'monster'
                          },
                          {
                            'q' => 'demon'
                          },
                          {
                            'q' => 'heller'
                          },
                          {
                            'q' => 'hellion'
                          },
                          {
                            'q' => 'demon'
                          },
                          {
                            'q' => 'daystar'
                          },
                          {
                            'q' => 'ogre'
                          },
                          {
                            'q' => 'tempter'
                          },
                          {
                            'q' => 'hell'
                          },
                          {
                            'q' => 'halloween'
                          }
                        ],
          'remove_duplicates' => \1
        };`

It *almost* works now. The counts still seem off though. 

`&sort_by=add_date:desc&max_extra_suffix=1&drop_tokens_threshold=1&query_by_weights=keywords:10,title:3,description:2,category:1&filter_by=nametag:=false&prefix=1&per_page=55&max_facet_values=100&exact_match_boost=1&page=47&prioritize_exact_match=1&facet_by=keywords_array&num_typos=1&max_extra_prefix=0&query_by=title,description,category,keywords`

I get back:

'found' => 2587,

So 2587 results, with 55 per page - gives me 47.036 pages. Yet on page 46, it finishes with just 26 results - and nothing on page 47.  So something still not quite right. Please let me know if you'd like some more debugging info

UPDATE: I just noticed, the total found changes depending on page:

```
page 41 -  2528
page 42 -  2491
page 43 -  2491
page 44 -  2523
page 45 -  2575
page 46 -  2,587
```

So it seems to fluctuate quite a lot, even with the same params being passed in (just a different page number) 

Cheers

Andy

happy-san:
@youradds If you can provide a minimal example [in this format](https://gist.github.com/jasonbosco/7c3432713216c378472f13e72246f46b), it would help us identify the issue. Feel free to open a new Github issue.

youradds:
> @youradds If you can provide a minimal example [in this format](https://gist.github.com/jasonbosco/7c3432713216c378472f13e72246f46b), it would help us identify the issue. Feel free to open a new Github issue.

Hi. Docker isn't really my language / method of use, but I've written a test case in Perl and put it here:

https://github.com/typesense/typesense/issues/2730

chatGPT gave me a converted script for bash/docker, but I can't guarantee it'll work. I've included the perl scripts, so hopefully you can use those to create a dataset and then test it with the same logic

PR #2576, opened 2025-09-22
Update SHOWCASE.md

kicaj:
- Otaku Square - domain is for sale


jasonbosco:
@kicaj Could you create a single PR based off of the v30 branch (not master), and do a PR to the v30 branch (which is the current default branch we're using)?

Issue #2559, opened 2025-09-12
Parser error "unbalanced `&&` operands" with valid filter query on collections with 78+ documents


danieltroger:
## Bug Description
Typesense returns a parser error "Could not parse the filter query: unbalanced `&&` operands" when executing a valid filter query with `&&` operator on a collection with 78+ documents and a complex schema (979 fields). The same query works fine with 77 documents.

## Reproduction Steps
We've created a minimal reproduction that reliably triggers this bug. Due to the complexity required (979-field schema + 78 documents), we've packaged it as a Python script rather than a curl one-liner.

**Reproduction repository (9.5 MB ZIP): [typesense-reproduction-minimal.zip](https://github.com/user-attachments/files/22303676/typesense-reproduction-minimal.zip)**

To reproduce:
```bash
# Extract the ZIP file
unzip typesense-reproduction.zip
cd typesense-reproduction

# Use the provided venv and run
source venv/bin/activate
pip install -r requirements.txt  # Dependencies may already be installed
python reproduce_bug.py
```

The equivalent curl command that triggers the bug (after loading the data):
```bash
curl -X POST 'http://localhost:9101/multi_search' \
  -H 'X-TYPESENSE-API-KEY: TYPESENSEDEV' \
  -H 'Content-Type: application/json' \
  -d '{
    "searches": [{
      "collection": "shopify_gpt_search_search-21143697__main_repro",
      "page": 1,
      "per_page": 1,
      "group_by": "product_id,variant_color__en",
      "group_limit": 10,
      "include_fields": "product_id,variant_id",
      "query_by": "",
      "q": "*",
      "vector_query": "embedding:([...1536 floats...], k:1)",
      "filter_by": "has_country_price__21061206196__SE:true && quantity_available:>0"
    }]
  }'
```

## Expected vs Actual
**Expected behavior**  
The query should execute successfully and return filtered results based on the `&&` condition.

**Actual behavior**  
Returns error: `Could not parse the filter query: unbalanced \`&&\` operands.`

## Environment
- **Typesense version:** v29.0 (tested on port 9101)
- **Operating system:** macOS
- **Client library & version:** typesense-python v1.1.1

## Schema / Configuration

The bug requires a complex schema with 979 fields. Key fields involved in the query:
```json
{
  "name": "shopify_gpt_search_search-21143697__main_repro",
  "fields": [
    { "name": "product_id", "type": "int32", "facet": true },
    { "name": "variant_color__en", "type": "string", "facet": true },
[...]

PR #2589, opened 2025-09-30
fix: remove trailing operands from filter query when group limit

ozanarmagan:
## Change Summary
Filter by clause has redundant trailing operands when one of the fields in the `groupBy` is optional and doesn't have values when group limit applied. 

## PR Checklist


ozanarmagan:
Fix for #2559

happy-san (approved):

Issue #2585, opened 2025-09-26
[Question] Gaps in Stemming Dictionary API and Documentation


bfritscher:
**Area:** API, Documentation, SDK

There are three key issues with the Stemming Dictionary API that make management difficult, stemming from discrepancies between observed behavior, SDK naming, and the official documentation.

**1. `import` Endpoint Duplicates Entries, Contradicting `upsert` in SDK**
The `POST /stemming/dictionaries/import?id=` endpoint (named `upsert` in the JS SDK) does not update existing entries. Importing a new `alternate` for an existing `root` creates a duplicate entry instead of overwriting the original one. This makes the `upsert` method name in the SDK misleading.

**2. No Method to Manage Individual Dictionary Entries**
The documentation provides no API endpoint or SDK method to delete or modify a single `root`/`alternate` pair within a dictionary. The only way to correct an error is to delete and recreate the entire dictionary.

**3. Dictionary `DELETE` Endpoint is Undocumented and Missing from SDKs**
The documentation for "Managing Dictionaries" is missing a `DELETE` endpoint. A `DELETE /stemming/dictionaries/{id}` route appears to exist in the source code but is not officially documented or implemented in the client SDKs.

### Questions

1.  Is the append-only behavior of the `import` endpoint intentional?
2.  Are there plans to add functionality for deleting or modifying individual stemming pairs?
3.  Can the `DELETE /stemming/dictionaries/{id}` endpoint be officially documented and added to the SDKs?

kishorenc:
1. Definitely a bug. We will fix it.
2. We don't have plans for this yet. We currently treat the whole dictionary as a single resource you manage. 
3. This should be documented.

PR #2594, opened 2025-10-06
fix upsert bug with stemming dictionary

krunal1313:
#### import Endpoint Duplicates Entries, Contradicting upsert in SDK
The `POST /stemming/dictionaries/import?id=` endpoint (named upsert in the JS SDK) does not update existing entries. Importing a new alternate for an existing root creates a duplicate entry instead of overwriting the original one. This makes the upsert method name in the SDK misleading.

### Fix
Update the root if word is already in the dictionary else emplace new word-root to dictionary

## PR Checklist


krunal1313:
Fixes #2585

kishorenc (approved):

PR #2655, opened 2025-11-03
fix(schema): only drop nested children when dropping object field

### LINKED ISSUE #2654: Dropping an object field also drops fields with a prefix of the field's name
## Bug Description

When dropping a nested object field, Typesense removes top-level fields that share the same prefix name. For example, dropping the `attributes` object field also removes unrelated top-level fields like `attributes_filter` and `attributes_nested_string`, even though they are not nested children of `attributes`.

## Reproduction Steps

```bash
export TYPESENSE_API_KEY=xyz
export PORT=8108
export TYPESENSE_HOST=http://localhost:$PORT

# Create collection with nested object field and top-level fields with similar prefix
curl "http://localhost:$PORT/collections" \
  -X POST \
  -d '{
    "name": "companies",
    "fields": [
      {"name": "attributes", "type": "object"},
      {"name": "attributes.filter", "type": "int64"},
      {"name": "attributes.nested_string", "type": "string"},
      {"name": "attributes_filter", "type": "int64"},
      {"name": "attributes_nested_string", "type": "string"}
    ],
    "enable_nested_fields": true
  }'

# Verify all fields exist
curl "http://localhost:$PORT/collections/companies" \
  -X GET \
  -H "X-TYPESENSE-API-KEY: ${TYPESENSE_API_KEY}" | jq '.fields[].name'

# Drop the attributes object field
curl "http://localhost:$PORT/collections/companies" \
  -X PATCH \
  -d '{
    "fields": [
      {"name": "attributes", "drop": true}
    ]
  }' | jq '.'

# Check remaining fields - attributes_filter and attributes_nested_string are incorrectly removed
curl "http://localhost:$PORT/collections/companies" \
  -X GET \
  -H "X-TYPESENSE-API-KEY: ${TYPESENSE_API_KEY}" | jq '.fields[].name'
```

## Expected vs Actual

**Expected behavior**  

When dropping the `attributes` object field, only the following should be removed:
- `attributes` (the object field itself)
- `attributes.filter` (nested child)
- `attributes.nested_string` (nested child)

The top-level fields `attributes_filter` and `attributes_nested_string` should remain intact since they are not nested children of `attributes`.

**Actual behavior**  

All fields with names starting with `attributes` are removed, including:
- `attributes` (correctly removed)
- `attributes.filter` (correctly removed)
- `attributes.nested_string` (correctly removed)
- `attributes_filter` (incorrectly removed)
- `attributes_nested_string` (incorrectly removed)

## Environment

- **Typesense version:** v30 (branch)
- **Operating system:** Linux 6.12.43-1-lts  (EndeavourOS)
- **Client library & version:** curl (direct API calls)

## Schema / Configuration

```json
{
  "name": "companies",
  "enable_nested_fields": true,
  "fields": [
    {"name": "attributes", "type": "object"},
    {"name": "attributes.filter", "type": "int64"},
    {"name": "attributes.nested_string", "type": "string"},
    {"name": "attributes_filter", "type": "int64"},
    {"name": "attributes_nested_string", "type": "string"}
  ]
}
```

## Additional Context

The bug is in the `validate_alter_payload` function in `src/collection.cpp`. When dropping a nested object field, the code uses `equal_prefix_range(field_name)` which matches any field starting with the field name. This causes false positives for top-level fields that happen to share the same prefix.

tharropoulos:
## Change Summary
when dropping a nested object field, prefix matching was incorrectly
removing top-level fields with similar names. fix by using `field_name + "."`
prefix to only match dot-notation nested children.

fixes #2654 

## PR Checklist

### DIFF
--- a/src/collection.cpp
+++ b/src/collection.cpp
@@ -6820,17 +6820,14 @@ Option<bool> Collection::validate_alter_payload(nlohmann::json& schema_changes,
 
                 // should also remove children if the field being dropped is an object
                 if(field_it.value().nested && enable_nested_fields) {
-                    auto prefix_it = search_schema.equal_prefix_range(field_name);
+                    auto prefix_it = search_schema.equal_prefix_range(field_name + ".");
                     for(auto prefix_kv = prefix_it.first; prefix_kv != prefix_it.second; ++prefix_kv) {
-                        bool exact_key_match = (prefix_kv.key().size() == field_name.size());
-                        if(!exact_key_match) {
-                            del_fields.push_back(prefix_kv.value());
-                            updated_search_schema.erase(prefix_kv.key());
-                            updated_nested_fields.erase(prefix_kv.key());
+                        del_fields.push_back(prefix_kv.value());
+                        updated_search_schema.erase(prefix_kv.key());
+                        updated_nested_fields.erase(prefix_kv.key());
 
-                            if(prefix_kv.value().embed.count(fields::from) != 0) {
-                                updated_embedding_fields.erase(prefix_kv.key());
-                            }
+                        if(prefix_kv.value().embed.count(fields::from) != 0) {
+                            updated_embedding_fields.erase(prefix_kv.key());
                         }
                     }
                 }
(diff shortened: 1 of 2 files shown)


src/collection.cpp
```diff
@@ -6820,17 +6820,14 @@ Option<bool> Collection::validate_alter_payload(nlohmann::json& schema_changes,
 
                 // should also remove children if the field being dropped is an object
                 if(field_it.value().nested && enable_nested_fields) {
-                    auto prefix_it = search_schema.equal_prefix_range(field_name);
+                    auto prefix_it = search_schema.equal_prefix_range(field_name + ".");
```
kishorenc, line 6823: @tharropoulos 

We are looking up `field_name + "."` here. How does this match the root object field in the schema? This prefix search should not match the value `attribute` (i.e. the parent object field name) right?

tharropoulos: The parent field is already deleted beforehand (on lines 6800-6802):
```cpp
            if(found_field) {
                del_fields.push_back(field_it.value());
                updated_search_schema.erase(field_it.key());
                updated_nested_fields.erase(field_it.key());
```

kishorenc (approved):

PR #2675, opened 2025-11-21
FEATURE: `POST /health {"shutdown":true}` sets failing health

pgray:
## Change Summary
this would add a `POST /health {"shutdown": true}` endpoint

in systems that determine traffic routing based on healthchecks (e.g. kubernetes),

typesense responding quickly to sigterm before healthchecks have time to fail results in traffic being routed to down nodes

this PR seeks to add a `shutdown` request that would mark the system as unhealthy but allow it to continue serving traffic still routed at it

## PR Checklist

### DIFF
--- a/src/core_api.cpp
+++ b/src/core_api.cpp
@@ -371,7 +371,7 @@ bool patch_update_collection(const std::shared_ptr<http_req>& req, const std::sh
             res->set_400("The `curation_sets` value should be an array.");
             return false;
         }
-        
+
         auto curation_sets = req_json["curation_sets"].get<std::vector<std::string>>();
         auto op = CollectionManager::get_instance().update_collection_curation_sets(req->params["collection"], curation_sets);
         if(!op.ok()) {
@@ -444,7 +444,7 @@ bool get_debug(const std::shared_ptr<http_req>& req, const std::shared_ptr<http_
 
 bool get_health_with_resource_usage(const std::shared_ptr<http_req>& req, const std::shared_ptr<http_res>& res) {
     nlohmann::json result;
-    bool alive = server->is_alive();
+    bool alive = server->is_alive() && !server->unhealthy_set();
 
     auto resource_check = cached_resource_stat_t::get_instance().has_enough_resources(
         Config::get_instance().get_data_dir(),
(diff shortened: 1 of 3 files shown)


kishorenc:
Would you be able to simply remove the pod from Kubernetes instead of making this API call to Typesense? I'm not sure whether we should introduce features that are catering to specific deployment patterns and their quirks.

pgray:
Any system that uses health checks as a way to determine which node to route traffic to will suffer from this issue. This makes typesense friendlier to being load balanced in general by allowing the operator to take it out of traffic with a "soft signal" and then later issuing sigterm

- [haproxy](https://www.haproxy.com/documentation/haproxy-configuration-tutorials/reliability/health-checks/#http-health-checks)
- [nginx](https://docs.nginx.com/nginx/admin-guide/load-balancer/http-health-check/#specifying-the-requested-uri)
- [envoy](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/health_check.proto#envoy-v3-api-msg-config-core-v3-healthcheck-httphealthcheck)

src/core_api.cpp
```diff
@@ -509,7 +509,30 @@ bool get_health(const std::shared_ptr<http_req>& req, const std::shared_ptr<http
...
+        } catch(const std::exception& e) {
+            // return 400 with the error message
+            // but also return the alive status
+            LOG(ERROR) << "JSON error: " << e.what();
+            result["ok"] = alive;
+            result["error"] = "Bad JSON.";
+            res->set_body(400, result.dump());
+            return alive;
+        }
+        if(req_json.contains("shutdown") && req_json["shutdown"] == true) {
```
pgray, line 527: is this right?

kishorenc:
Ok let me summarize my understanding of the issue you are facing and suggest an alternative which I think is better.

The problem (as I understand):

1. When Typesense receives SIGTERM (e.g., during a Kubernetes pod termination), it shuts down quickly
2. But the load balancer's health checks haven't had time to detect the failure yet
3. Traffic continues to be routed to a node that's already dead or dying, causing errors

Is this correct?

Instead of overload `/health` we can do this:

1. Immediately start failing health checks (return 503 on /health)
2. Continue serving in-flight requests for a configurable period (e.g., --shutdown-delay-seconds=30)
3. Then terminate cleanly

This approach requires no external coordination or API calls and works with any load balancer that uses health checks.
Nginx uses this pattern via the `worker_shutdown_timeout` directive.

pgray:
That would certainly be workable for my use-case.

Thanks so much for the response.

PR #2698, opened 2025-12-04
Clean up entries in SHOWCASE.md

kicaj:
Removed deprecated links entries and cleaned up the showcase list.


jasonbosco:
Thank you!

Issue #2697, opened 2025-12-03
[Feature Request] Support for synonym matches in curation rule


jasonbosco:
Let's say there's a curation rule where `rule.query` is set to `abc`.

Independently, let's say `abc` and `xyz` are defined as synonyms of each other.

Now when a user searches for `xyz`, we want to trigger the curation rule above defined for `abc`, since it's a synonym of `abc`. 

We want this behavior to be managed via a flag in the curation rule, called `rule.synonyms: true | false`

alangmartini:
Available in v30

PR #2708, opened 2025-12-12
Support synonyms with curations

krunal1313:
### Synonyms matching with Curations
To enable synonyms matching with curations, one need to set `synonyms:true` in curation rule like below,
```json
{
        "id": "synonyms2",
        "rule": {
            "query": "credit card",
            "match": "exact",
            "synonyms" : true
          },
          "includes": [
                {"id": "4", "position": 1}
         ]
}
```
With that enabled, if `query` tokens are not matching any curation rule, it'll try to match curations using synonyms of query tokens.
For e.g
Suppose there is synonym added to collection,
```json
{
        "id": "card-synonyms",
        "synonyms": ["credit card", "payment card", "cc"]
}
```
Now if we search the collection using query `"payment card"` or `"cc"`, it'll trigger the curation defined above as both are synonym of `credit card`.

## PR Checklist

### DIFF
--- a/include/collection.h
+++ b/include/collection.h
@@ -533,7 +533,8 @@ private:
                              std::vector<const curation_t*>& filter_curations,
                              bool& filter_curated_hits,
                              std::string& curated_sort_by,
-                             nlohmann::json& curation_metadata) const;
+                             nlohmann::json& curation_metadata,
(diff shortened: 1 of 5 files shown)


kishorenc (approved):

kishorenc:
Fixes https://github.com/typesense/typesense/issues/2697

PR #2711, opened 2025-12-17
add param shutdown-delay-seconds

krunal1313:
### Delayed Shutdown
One can set the param `shutdown-delay-seconds=<seconds>` as command line param or put in config file to set the delayed shutdown seconds.
#### purpose
- in case of terminating server, any in-flight ongoing requests going on gets terminated immediately.
- with the above param, system gets some time to finish up inflight on going requests before shutting down.

#### behavior
- One need to send `SIGHUP` signal to typesense-server for triggering this delayed shutdown countdown. After which typesense server will be shut.
- after successfully triggering shutdown, /health endpoint will be returning error `503`.
- in-flight requests will not be terminated and gets time till the timeout.
- one can adjust delayed seconds as per their need.
- other signals like `SIGINT`, `SIGTERM` dont handle above param and will behave as usual.

## PR Checklist

### DIFF
--- a/include/http_server.h
+++ b/include/http_server.h
@@ -162,6 +162,8 @@ private:
 
     ThreadPool* meta_thread_pool;
 
+    bool is_shutdown_triggered = false;
+
     bool (*auth_handler)(std::map<std::string, std::string>& params,
                          std::vector<nlohmann::json>& embedded_params_vec,
                          const std::string& body, const route_path& rpath,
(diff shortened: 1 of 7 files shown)


krunal1313:
Supersedes #2675

pgray:
:heart:

pgray:
would love to see this in a `v0.30.1` or something 🙏

kishorenc (approved):

pgray:
woooooooooooooooooooooooooooooo

PR #2721, opened 2025-12-24
fix: include num_dim in remote embedder cache key for matryoshka models

onceLearner:
### Remote Embedder Cache Key Bug explained:
When using `matryoshka-compatible` embedding models (e.g., `text-embedding-3-large`, `text-embedding-3-small`) with different `num_dim` values across multiple _collections_ or _fields_, __all fields incorrectly share the same cached embedder instance__.

 This causes embeddings to be generated with the wrong dimensions => Wrong Cosine/dot distances.
 
For example:
Create field A with num_dim: 768 using openai/text-embedding-3-large
Create field B with num_dim: 512 using the same model
Field B's embeddings are incorrectly generated with 768 dimensions
Root Cause:
The embedder cache key was generated using only model_name:api_key, ignoring the requested dimensions:

```cpp
// src/text_embedder_remote.cpp
std::string OpenAIEmbedder::get_model_key(..)
return model_config["model_name"] + ":" + model_config["api_key"];
```

### Changes:

- Added optional `num_dims` parameter (default 0) to `get_model_key()` for all remote embedders
- Updated `get_text_embedder()` to accept and propagate `num_dims`
- Updated all call sites to pass the field's `num_dim`
- Added unit test to verify cache key behavior

## PR Checklist

### DIFF
--- a/include/embedder_manager.h
+++ b/include/embedder_manager.h
@@ -42,7 +42,7 @@ public:
     EmbedderManager(const EmbedderManager&) = delete;
     EmbedderManager& operator=(const EmbedderManager&) = delete;
 
-    Option<TextEmbedder*> get_text_embedder(const nlohmann::json& model_config);
+    Option<TextEmbedder*> get_text_embedder(const nlohmann::json& model_config, size_t num_dims = 0);
     Option<ImageEmbedder*> get_image_embedder(const nlohmann::json& model_config);
 
     void delete_text_embedder(const std::string& model_path);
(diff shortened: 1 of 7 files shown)


tharropoulos:
Hey @onceLearner,

I've updated the PR on #2724. We'll take the changes on from there

onceLearner:
> Hey @onceLearner,
> 
> I've updated the PR on #2724. We'll take the changes on from there

Confirm It is fixed ! thanks so much team!

Issue #2732, opened 2026-01-05
Latency with overflowed numbers.


alangmartini:
## Bug Description
On v30.0.rc30, metrics on /stats.json are showing outrageously big numbers for latency, which don't reflect the real latency.

```
{
  "cache_hit_ratio": 0.0,
  "delete_latency_ms": 0,
  "delete_requests_per_second": 0,
  "import_latency_ms": 0,
  "import_requests_per_second": 0,
  "latency_ms": {
    "GET /collections/events/documents/search": 12.090909090909092,
    "GET /health_with_rusage": 100.66666666666667,
    "GET /stats.json": 0.0
  },
  "overloaded_requests_per_second": 0,
  "pending_write_batches": 0,
  "requests_per_second": {
    "GET /collections/events/documents/search": 6.6,
    "GET /health_with_rusage": 1.5,
    "GET /stats.json": 0.1
  },
  "search_70Percentile_latency_ms": 18444978212771213000.0,
  "search_95Percentile_latency_ms": 18444978212772340000.0,
  "search_99Percentile_latency_ms": 18444978212772774000.0,
  "search_latency_ms": 279496122328932600.0,
  "search_max_latency_ms": 18444978212772773888,
  "search_min_latency_ms": 18444978212762861568,
  "search_requests_per_second": 6.6,
  "total_requests_per_second": 8.2,
  "write_latency_ms": 0,
  "write_requests_per_second": 0
}
```
## Environment
- **Typesense version:** e.g. `v30.0.rca30`

krockxz:
Hey @alangmartini 
I have analysed the issue and looks like the issue is caused by using inconsistent clock sources resulting in an integer underflow

kishorenc:
Fixed in v30.1

PR #2733, opened 2026-01-06
Fix latency metrics overflow issue #2732

krockxz:
## Change Summary
Issue
The /stats.json endpoint was reporting invalid, extremely large values (e.g., 1.84e19) for latency metrics. This was caused by a mismatch in clock sources and units in CollectionManager::do_search.

>Start Time: Calculated using std::chrono::system_clock (in http_req constructor).
>End Time: Calculated using std::chrono::high_resolution_clock (in do_search).
>Unit Mismatch: The subtraction was checking different units without proper conversion.

Fix:
Updated src/collection_manager.cpp to use std::chrono::system_clock for the end time, ensuring it matches the start time source. Both are now converted to microseconds before subtraction, then converted to milliseconds.
 
 Verification
Added a regression test VerifySearchLatencyMetric in 
test/collection_manager_test.cpp

>Simulates a search request with a known start time.
>Verifies that search_latency_ms in AppMetrics is non-negative and within reasonable bounds (< 100s).

## PR Checklist

- [x ] I have read and signed the [Contributor License Agreement](https://forms.gle/PZyiY5N2GDQU8GsV9).

### DIFF
--- a/test/app_metrics_test.cpp
+++ b/test/app_metrics_test.cpp
@@ -1,16 +1,35 @@
 #include <gtest/gtest.h>
+#include <string>
+#include <vector>
+#include <collection_manager.h>
 #include "app_metrics.h"
+#include "collection.h"
 
 class AppMetricsTest : public ::testing::Test {
 protected:
     AppMetrics& metrics = AppMetrics::get_instance();
+    Store *store;
+    CollectionManager & collectionManager = CollectionManager::get_instance();
+    std::atomic<bool> quit = false;
 
     virtual void SetUp() {
-      metrics.window_reset();
+        std::string state_dir_path = "/tmp/typesense_test/app_metrics_test_db";
+        system(("rm -rf "+state_dir_path+" && mkdir -p "+state_dir_path).c_str());
+
+        store = new Store(state_dir_path);
+        collectionManager.init(store, 1.0, "auth_key", quit);
+        collectionManager.load(8, 1000);
+        
+        metrics.window_reset();
     }
 
     virtual void TearDown() {
-      metrics.window_reset();
+        metrics.window_reset();
+        
+        if(store != nullptr) {
+            collectionManager.dispose();
+            delete store;
+        }
     }
 };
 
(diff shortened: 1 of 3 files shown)


krunal1313:
@krockxz ,
union search case is not handled in PR.
can you update it?

krockxz:
WIP

krockxz:
@krunal1313 Updated

test/collection_manager_test.cpp
```diff
@@ -2104,4 +2105,111 @@ TEST_F(CollectionManagerTest, CloneCollectionWithDocuments) {
...
+    ASSERT_TRUE(metrics_result.contains("search_latency"));
+    double latency = metrics_result["search_latency"];
+    
+    // Check for overflow (huge values) or negative values
+    // The bug produced values like 1.84e19
+    ASSERT_GE(latency, 0.0);
+    ASSERT_LT(latency, 100000.0); // 100s is plenty margin, well below 1.84e19
+
+    collectionManager.drop_collection("union_latency_test1");
+    collectionManager.drop_collection("union_latency_test2");
```
krunal1313: Move these tests to app_metrics_test.cpp

you can add test for multi search without union too and name them better way like,
1) SearchLatencyStats - normal search latency stats
2)  MultiSearchLatencyStats - multi search latency stats (with and without union)

krockxz: Made the changes, take a look

krockxz: Made the changes, take a look

test/app_metrics_test.cpp
```diff
@@ -82,4 +101,133 @@ TEST_F(AppMetricsTest, EstimateQuantileDuration) {
...
+    // Test 1: Multi-search WITHOUT union
+    {
+        metrics.window_reset();
+
+        auto start_ts = std::chrono::duration_cast<std::chrono::microseconds>(
+                std::chrono::system_clock::now().time_since_epoch()).count();
+
+        std::this_thread::sleep_for(std::chrono::milliseconds(10));
+
+        auto multi_op = collectionManager.do_union(req_params, embedded_params_vec, searches, response, start_ts, false);
```
krunal1313: this is not how without union multi-search works.
do_union() is explicitly called for union searches.
check post_multi_search(req, res) for better understanding how multi-searches work.

krockxz: Updated

krunal1313 (approved):

krunal1313 (changes requested):
Tests are not strong enough.
tests pass without fix too.
update the tests.

krockxz:
> Tests are not strong enough. tests pass without fix too. update the tests.

tests are now checking 10ms <= latency < 1000ms
should work
@krunal1313

krunal1313:
@krockxz not really.
try it yourself.
check locally thoroughly and then commit. Tests should not pass without fix.

krockxz:
> @krockxz not really. 

I ran the most recent committed tests (commit 5efcb99d) with the older buggy code on my system, and all three tests fail as expected
All three tests show latency = 1.8e19 ms, which correctly fails the ASSERT_LT(latency, 1000.0) assertion

To debug the discrepancy, could you share:

Which specific tests are passing?
What latency value you're seeing?
Your OS, compiler version, and architecture
The exact test command used

The bug mixes milliseconds with microseconds, producing huge values. If tests pass on your system it could be clock behavior differences or  compiler optimizations or timing-specific results which I am currently unsure of

krunal1313:
@krockxz, tests need to be updated with fix too.

ealier tests were always passing because condition was `ASSERT_GE(latency, 0.0)`.
In the tests, latency will come as 0 always as it'll not compute latency until window is reset. You can check `app_metrics.cpp` for better understanding.
If latest tests are passing in your machine with fix, then it is wrong.

kishorenc:
Thanks for the PR @krockxz -- I have merged the patch for this change without the change. The test has a logical flaw because it adds a sleep in the test but expecting the returned latency metric to increase. This won't happen and it's a bit difficult to reproduce in a test.

Issue #2717, opened 2025-12-19
[Feature Request] Curate by filter_by needs to parse the filter expression to do a match


alangmartini:
We currently do a simple exact match between the `filter_by` mentioned in the curation rule vs the search query. 

The problem with this approach is that the following rule, for eg, does not get triggered for any of the search queries below that, even though they are all functionally equivalent:

Curation Rule: `rule.filter_by = field:[abc, xyz]`

Search Queries that should trigger the rule above, but currently don't: 
- ```filter_by = field:[`abc`, `xyz`]``` (notice the backticks used to escape)
- `filter_by = field:[xyz, abc]` (notice the order of terms is different, but it's still functionally the same)
- `filter_by = field:xyz || field:abc` (same effective operation, but using different syntax)
- `filter_by = field:[abc, xyz, pqr]` (search query fully contains all the OR values mentioned in the curation rule, so it should still trigger the curation rule)

Here's another example:

Curation Rule: `rule.filter_by = field1:[abc, xyz] && field2:pqr`

Search Queries that should trigger the rule above, but currently don't: 
- `rule.filter_by = field2:pqr && field1:[abc, xyz]`
- `rule.filter_by = field2:pqr && field1:[xyz, abc]`
- ```rule.filter_by = field2:`pqr` && field1:[xyz, abc]```

The solution to this would be to parse `filter_by` expression in the curation rule and then compare the parsed expression with the parsed filter in the search query.

## Reproduction Steps
Please include a minimal, reproducible example **using curl**. [Use this script as a guide.](https://gist.github.com/jasonbosco/7c3432713216c378472f13e72246f46b)

```bash
#!/bin/bash

### Reproduction Script: Curation Rules filter_by Matching Bug ###
# Bug: Curation rules only trigger when filter_by clause exactly matches the string defined.
# - filter_by=is_available:true won't match filter_by=is_available:[true]
# - filter_by=is_available:true && is_enabled:true won't match reversed order

### Run Typesense via Docker ########################################
export TYPESENSE_API_KEY=xyz
export PORT=8108
export TYPESENSE_HOST=http://localhost:$PORT

docker stop typesense-repro 2>/dev/null
docker rm typesense-repro 2>/dev/null
rm -rf "$(pwd)"/typesense-data-dir-repro
mkdir "$(pwd)"/typesense-data-dir-repro

# Start Typesense
docker run -d -p $PORT:$PORT --name typesense-repro \
[...]

PR #2757, opened 2026-02-03
semantic match for curation filter_by

krunal1313:
## PR Checklist

- Supports semantic match for curation filter_by instead of simple string match
     Fixes following smentics currently

     - Order of values in lists
     - Backticks / quoting variations
     - Commutativity of &&
     - Different syntax for OR (|| vs […])
     - Superset matching (query has extra values)
     - Basic numeric comparisons and ranges
     - Logical grouping with parentheses
     
- Added sample tests


happy-san:
https://github.com/typesense/typesense/issues/2717

PR #2768, opened 2026-02-08
fix(filter): correct deep nested field validation in delete operations

### LINKED ISSUE #2765: API Key for operations/schema_changes not working
## Bug Description
I tried to create an API Key, that is allowed to get schema_changes.

## Reproduction Steps
Take the action from https://typesense.org/docs/30.0/api/api-keys.html#create-an-api-key and create an api key
```yaml
{
  "description": "test",
  "actions": [
    "operations/schema_changes:get"
  ],
  "collections": [
    "*"
  ]
}
```
Call GET /operations/schema_changes with this API key

## Expected vs Actual
**Expected behavior**  
HTTP 200 response

**Actual behavior**  
HTTP 401 Forbidden

## Environment
- **Typesense version: ** `v29.0`
- **Operating system:** `Ubuntu 22.04 x86_64`

JosephDoUrden:
## Summary

Fixes incorrect document deletion when using deep nested field filters
(e.g. `main.items.{status: active && price: <100}`).

---

## Root Cause

`validate_object_filter()` in `src/filter_result_iterator.cpp` has two code paths:

- Batched path (when `is_filter_result_initialized` is true)
- Non-batched path (used during iterative AND/OR evaluation)

The batched path correctly used `get_nested_field_doc()` to traverse deep
nested JSON structures.

However, the non-batched path accessed:

document[filter_node->object_field_name]

This fails for multi-level paths such as `main.items`, because it attempts
to look up the literal key instead of navigating:

main → items

As a result, nested filters could match incorrectly, leading to unintended
document deletions.

---

## Solution

Updated the non-batched path to use `get_nested_field_doc()`, making it
consistent with the batched implementation.

### Before

for (const auto& nested_object: document[filter_node->object_field_name])

### After

const auto& doc = get_nested_field_doc(filter_node->object_field_name, document);
for (const auto& nested_object: doc)

---

## Files Changed

- src/filter_result_iterator.cpp  
  → Use get_nested_field_doc() in non-batched validation path (1-line fix)

- test/core_api_utils_test.cpp  
  → Added two test cases covering nested and deep-nested delete filters

---

## Test Plan

Added:

- DeleteDocumentsWithNestedFieldFilter
- DeleteDocumentsWithDeepNestedFieldFilter

Both tests verify:

- Only matching documents are deleted
- Non-matching documents remain
- AND conditions within the same nested object are respected

Run tests:

bazel test //:typesense-test --test_filter="CoreAPIUtilsTest.DeleteDocumentsWithNestedFieldFilter"

bazel test //:typesense-test --test_filter="CoreAPIUtilsTest.DeleteDocumentsWithDeepNestedFieldFilter"

Or full suite:

bazel test //:typesense-test

---

## Checklist

- [x] Minimal fix (single logic change)
- [x] Tests added
- [x] No unrelated refactors

Closes #2765

### DIFF
--- a/src/filter_result_iterator.cpp
+++ b/src/filter_result_iterator.cpp
@@ -3301,7 +3301,8 @@ bool filter_result_iterator_t::validate_object_filter() {
         return false;
     }
 
-    for (const auto& nested_object: document[filter_node->object_field_name]) {
+    const auto& doc = get_nested_field_doc(filter_node->object_field_name, document);
+    for (const auto& nested_object: doc) {
         if (validate_object_filter_helper(index, nested_object, filter_node)) {
             return true;
         }
--- a/test/core_api_utils_test.cpp
+++ b/test/core_api_utils_test.cpp
@@ -3120,3 +3120,173 @@ TEST_F(CoreAPIUtilsTest, UnionRemoveDuplicates) {
     ASSERT_EQ("1", response["hits"][3]["document"]["id"]);
     ASSERT_EQ("1", response["hits"][4]["document"]["id"]);
 }
+
+TEST_F(CoreAPIUtilsTest, DeleteDocumentsWithNestedFieldFilter) {
+    // Test delete with nested object filter to ensure only intended documents are deleted
+    nlohmann::json schema_json = R"({
+        "name": "nested_delete_test",
+        "fields": [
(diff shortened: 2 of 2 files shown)


kishorenc:
The PR is not relevant to linked issue.

PR #2772, opened 2026-02-11
fix: route /health endpoint to meta_thread_pool to prevent timeouts during bulk inserts

### LINKED ISSUE #2638: High write queue during bulk inserts blocks /health, /search, and /insert APIs (server becomes unresponsive)
# Typesense Unresponsiveness During Large Batch Insert Operations

## Description

Hi team 👋,

I'm encountering a consistent issue where Typesense becomes completely unresponsive (including `/health` and `/search` endpoints) during large batch insert operations — even though the process eventually completes or fails gracefully.

## Environment

- **Typesense version**: v0.25.2
- **Deployment**: Docker container (single node)
- **Machine spec**: 2 vCPUs, 2 GB RAM, 30 GB disk (Ubuntu 22.04)
- **Collection size**: ~228,000 documents
- **Schema highlights**:
  - 512-dimension embedding vector field (float[])
  - Several indexed string fields (title, description, etc.)
  - Multiple facet and numeric fields

## Observed Behavior

After inserting ~200K documents successfully, the next bulk insert (10K records batch) triggers these logs repeatedly:
```
E raft_server.cpp:771] 4663 queued writes > healthy read lag of 1000
E raft_server.cpp:783] 4663 queued writes > healthy write lag of 500
I raft_server.cpp:692] Term: 7, pending_queue: 0, last_index: 11852, committed: 11852, known_applied: 11852, applying: 0, pending_writes: 0, queued_writes: 4642, local_sequence: 664063
```

From this point:

- `/health` endpoint stops responding (timeout).
- `/search` and `/collections/.../documents` also hang indefinitely.
- Even simple GET requests are delayed or blocked until queued writes drain.
- The only fix is restarting the container, which releases the queue but loses pending writes.

## Expected Behavior

Even under high indexing load:

- The `/health` API should remain responsive.
- Read/search requests could degrade gracefully (e.g., limited throughput or temporary "overloaded" responses) rather than fully blocking.
- The server should handle backpressure more predictably — perhaps by rejecting new write batches early instead of letting the Raft write queue grow beyond recovery.

## Steps to Reproduce

1. Create a collection similar to:
```json
{
  "name": "bench_jobs",
  "fields": [
    {"name": "description", "type": "string"},
    {"name": "embedding", "type": "float[]", "num_dim": 512, "vec_dist": "cosine"}
  ],
  "default_sorting_field": "modified_at"
}
```

2. Insert ~200K documents in batches of 1000.
3. Continue inserting another 100K+ documents.
4. Observe server logs — queued writes rise beyond 3000–4000.
5. Try calling `/health` or `/search` — both timeout.

## Diagnostics

- System memory stays below total capacity (no OOM).
- CPU usage spikes, but container remains alive.
- Swap memory present but seemingly unused by Typesense.
- Disk space remains >50% free.

## Possible Cause (Hypothesis)

It seems that the internal Raft replication / write queue (`queued_writes`) is blocking the main event loop or thread pool used by HTTP handlers.

As a result, when large batches back up, the Raft thread starves the HTTP server threads, leading to total unresponsiveness.

This might be mitigated by:

- Decoupling `/health` and read routes from Raft's main write loop.
- Introducing configurable backpressure (reject or throttle writes before saturation).
- Allowing Typesense to drop queued writes more gracefully, rather than fully locking the node.

## Impact

- Makes large-scale indexing impractical on lower-memory instances.
- Prevents using Typesense in production ingestion pipelines that rely on health checks for availability.
- Causes system downtime during large bulk operations.

## Request / Suggestion

Could you please consider:

- Making `/health` endpoint independent of Raft queue state (non-blocking).
- Introducing a configurable write backpressure limit that rejects new inserts early with a clear 429/503 error.
- Providing visibility metrics (e.g., `/metrics` or `/debug`) to monitor Raft queue depth and pending writes in real-time.

## Additional Notes

Happy to provide:

- Collection schema
- Docker compose setup
- Exact load script (Python client)

if needed for reproducing.

Thanks for all your work — Typesense has been great overall, but this blocking behavior under load feels like a critical edge case worth improving. 🙏

MutharasuArchunan13:
## Summary
- Route `/health` requests to the dedicated `meta_thread_pool` (4 threads) instead of the main `thread_pool`
- Ensures `/health` stays responsive even when the main pool is saturated by bulk imports

## Fixes
Closes #2638

## Change
One-line change in `src/http_server.cpp`:
```cpp
// Before
bool use_meta_thread_pool = (root_resource == "status");

// After
bool use_meta_thread_pool = (root_resource == "status" || root_resource == "health");
```

## Why this works
- `meta_thread_pool` is a separate `ThreadPool(4)` never used for writes, searches, or imports
- `/health` is already a lightweight liveness probe (bypasses replication readiness checks)
- `/status` already uses this pool for the same isolation reason

### DIFF
--- a/src/http_server.cpp
+++ b/src/http_server.cpp
@@ -496,7 +496,7 @@ int HttpServer::catch_all_handler(h2o_handler_t *_h2o_handler, h2o_req_t *req) {
              root_resource == "config" || root_resource == "status" || root_resource == "proxy_sse"
          );
 
-    bool use_meta_thread_pool = (root_resource == "status");
+    bool use_meta_thread_pool = (root_resource == "status" || root_resource == "health");
 
     if(needs_readiness_check) {
         bool write_op = is_write_request(root_resource, http_method, rpath->handler);


suchith83:
@MutharasuArchunan13  Good observation. I also faced this issue. When I try to access write requrest more than million times. then health check is blocked. due to which I couldn't check if my application is running or not. @sunny can you merge this pr.

sunny:
Can’t, because I’m not a maintainer ;)

kishorenc (approved):

Issue #2753, opened 2026-01-30
[Feature Request] Delete document during bulk import


millimoose:
**Describe your feature request**
It should be possible to send "tombstone" documents to the bulk import API to delete documents with those IDs.

**Please describe your use-case for the feature request**
I'm trying to build a streaming pipeline to transparently update a search index and also ditch Algolia for Typesense. In my proof-of-concept, I feed typesense from a lambda that drains a Kafka topic with pending index updates. This kafka topic has a mix of messages that represent document upserts, as well as document deletions ("tombstone" messages where the value is empty). It would be nice if I could just send something like `{ id: 123, $deleted: true }` as one of the items of the bulk import endpoint request and be able to process this batch.

**Describe alternatives you've considered**
Currently I have to take batches of messages, split them into runs of upserts and tombstones, and process there runs sequentially. This means I all but have to use code to import documents and deal with a consumer process, as opposed to just feeding my topic into some fairly trivial Kafka Connect pipeline and be done with it.

PR #2776, opened 2026-02-13
batch import per doc action type

krunal1313:
### `all` action type
Typesense currently supports create, upsert, update actions type for importing docs.
These actions are defined at root level which applies to all docs in the batch.

#### Issues
In some pipelined jobs for adding, updating, deleting docs in the same batch one needs to filter and segregate all such docs and call their respective APIs for operation.

#### Proposed Approach
- Introduced a new per doc level approach to process doc on the go in same batch.
- new root level action type `all` is introduced to denote that import batch contain doc level actions which can be called like below,
```curl
curl -H "X-TYPESENSE-API-KEY: ${TYPESENSE_API_KEY}" \
      -X POST \
      -T documents.jsonl \
      "http://localhost:8108/collections/companies/documents/import?action=all"
```
and at doc level, one need to specify action like below
```json
{"id": "1", "company_name": "Stark Industries", "num_employees": 5215, "country": "USA", "$action": "create"}
{"id": "2", "company_name": "Orbit Inc.", "num_employees": 256, "country": "UK", "$action": "create"}
{"id": "1", "$action": "delete"}
```
`$action` value can be any of `create`, `update`, `delete`, `upsert`

In above sample import payload, in the end only doc with `id:2` will be indexed.

Similar way,
```json
{"id": "3", "company_name": "Avengers Inc", "num_employees": 5215, "country": "USA", "$action": "create"}
{"id": "4", "company_name": "Inifnity Inc.", "num_employees": 256, "country": "UK", "$action": "create"}
{"id": "4", "company_name": "Hulk Smashers", "$action": "update"}
```
for above import payload, doc with `id:3` will be indexed as usual but doc with `id:4` will be indexed with updated `company_name: Hulk Smashers`.

#### Notes
- when using `all` action type at rootlevel, at doc level  `$action` type is mandatory. Not doing so will  fail indexing  the doc.
- Normal uniform action types `create`, `update`, and `upsert` are still functional as before and can be used.

## PR Checklist

### DIFF
--- a/include/collection.h
+++ b/include/collection.h
@@ -25,6 +25,8 @@
 struct doc_seq_id_t {
     uint32_t seq_id;
     bool is_new;
+    index_operation_t operation;
+    bool is_same_batch = false;
 };
 
 struct highlight_field_t {
@@ -859,6 +861,7 @@ public:
     Option<doc_seq_id_t> to_doc(const std::string& json_str, nlohmann::json& document,
                                 const index_operation_t& operation,
                                 const DIRTY_VALUES dirty_values,
+                                const spp::sparse_hash_map<std::string, uint32_t>& batch_docs,
                                 const std::string& id="");
 
 
--- a/include/field.h
+++ b/include/field.h
@@ -464,8 +464,25 @@ enum index_operation_t {
     UPSERT,
     UPDATE,
     EMPLACE,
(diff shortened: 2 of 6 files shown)


krunal1313:
Fixes #2753

Issue #790, opened 2022-11-14
How to query for documents where an optional field is not set


Bug-Reaper:
How do you query for documents where an optional field is not set?

My apologies because I'm 99% sure I've seen this answered before but can't find the GH thread. I've searched for like an hour and can't find it. Probably would be good to get this added to the [docs](https://typesense.org/docs/0.23.1/api/search.html#query-parameters).

Bug-Reaper:
Thanks for the quick answer @jasonbosco. 

Since I already have a million+ records created, I'll handle it via ljson export of the entire db => script to do processing => re-upload of entire db. Slow, but it should get the job done.

Can I keep this open as a feature request?

jasonbosco:
Yeah, let's leave this open as a feature request. 

It might be faster to read the data from your primary datasource and then use the documents/import endpoint with `action= emplace` to just add a single field to your existing Typesense collection.

Bug-Reaper:
Re-discovered this today [sorting-null-empty-or-missing-values](https://typesense.org/docs/0.23.1/api/search.html#sorting-null-empty-or-missing-values). 

Doesn't fit my use-case but for anyone that wants to find all documents with unset optional numeric fields, you could use `filter_by` to arrange them at the top.

You _can_ also use `filter_by` to isolate documents with unset optional string fields. Not recommended as it can require a lot of extra memory, particularly for long-string fields like description.

Spovis:
Encountering this situation myself, would love to have this as a feature

elChapoSing:
Also keen on that feature !

shaunbyrnes:
As discussed on Slack, this should be introduced as a feature - doing something like IS NULL or colName != '' in SQL is very common and particularly for NoSQL DBs, which don't have as strict schemas, they can often have undefined/null/empty string fields, which would be expected behaviour and needs a way to filter on.

Look forward to having this added at some time :)

vladjerca:
My personal argument in adding placeholders from our side is that we end up loading junk into memory basically (maybe it's good for cloud sales 😜).

Having system level support to filter out entries with undefined or empty fields would be the optimal approach here.

PR #2777, opened 2026-02-14
feat: add existence filter for optional fields

ozanarmagan:
## Change Summary
Add support for filtering documents by whether optional fields are missing
using `field: _missing` and `field: !_missing` syntax.

This is enabled per field via the new `track_missing_values` schema property.

### Usage

1. Create a collection with an optional indexed field that has
   `track_missing_values` enabled:

```json
{
  "name": "products",
  "fields": [
    {"name": "color", "type": "string", "optional": true, "track_missing_values": true},
    {"name": "points", "type": "int32"}
  ]
}
```

2a. Filter for documents where the field is missing:

```http
GET /collections/products/documents/search?q=*&filter_by=color: _missing
```

2b. Filter for documents where the field is present:

```http
GET /collections/products/documents/search?q=*&filter_by=color: !_missing
```

2c. Combine with other filters:

```text
filter_by=color: _missing && points: >10
filter_by=color: !_missing || rating: _missing
```

## PR Checklist


ozanarmagan:
Fixes #790

test/collection_filtering_test.cpp
```diff
@@ -4258,4 +4258,445 @@ TEST_F(CollectionFilteringTest, DeepNestedObjectFieldsFiltering) {
...
+
+    auto op = collectionManager.create_collection(schema);
+    ASSERT_TRUE(op.ok());
+    auto coll = op.get();
+
+    ASSERT_TRUE(coll->add(R"({"id": "0", "title": "Shirt", "color": "red", "points": 10})"_json.dump()).ok());
+
+    auto search_op = coll->search("*", {}, "color: _exists", {}, sort_fields, {0});
+    ASSERT_FALSE(search_op.ok());
+    ASSERT_NE(std::string::npos, search_op.error().find("optional_index"));
```
happy-san: We should check against the actual error message.

test/collection_filtering_test.cpp
```diff
@@ -4258,4 +4258,445 @@ TEST_F(CollectionFilteringTest, DeepNestedObjectFieldsFiltering) {
...
+
+    search_op = coll->search("*", {}, "color: !_exists", {}, sort_fields, {0});
+    ASSERT_FALSE(search_op.ok());
+    ASSERT_NE(std::string::npos, search_op.error().find("optional_index"));
```
happy-san: We should check against the actual error message.

test/collection_filtering_test.cpp
```diff
@@ -4258,4 +4258,445 @@ TEST_F(CollectionFilteringTest, DeepNestedObjectFieldsFiltering) {
...
+    results = coll->search("*", {}, "color: !_exists && points: >10",
+                           {}, sort_fields, {0}).get();
+    ASSERT_EQ(1, results["found"].get<size_t>());
+    ASSERT_EQ("1", results["hits"][0]["document"]["id"].get<std::string>());
+
+    // OR
+    results = coll->search("*", {}, "color: _exists || points: >15",
+                           {}, sort_fields, {0}).get();
+    ASSERT_EQ(3, results["found"].get<size_t>());
+
```
happy-san, line 4485: We should add 
 ```c++
    std::vector<std::string> expected_ids = {"2", "1", "0"};
    for (size_t i = 0; i < results["hits"].size(); i++) {
        ASSERT_EQ(expected_ids[i], results["hits"][i]["document"]["id"].get<std::string>());
    }
```
to make sure only the expected ids match, in order.

File: test/collection_filtering_test.cpp [hunk 4]
```diff
@@ -4258,4 +4258,445 @@ TEST_F(CollectionFilteringTest, DeepNestedObjectFieldsFiltering) {
...
+    // text search + exists
+    results = coll->search("shirt", {"title"}, "color: _exists",
```
happy-san: `shirt` will match only one document regardless of the `filter_by` we should change this to `s` instead.

test/collection_filtering_test.cpp
```diff
@@ -4258,4 +4258,445 @@ TEST_F(CollectionFilteringTest, DeepNestedObjectFieldsFiltering) {
...
+    // both missing
+    results = coll->search("*", {}, "color: !_exists && size: !_exists",
+                           {}, sort_fields, {0}).get();
+    ASSERT_EQ(1, results["found"].get<size_t>());
+    ASSERT_EQ("3", results["hits"][0]["document"]["id"].get<std::string>());
+
+    // either exists (OR)
+    results = coll->search("*", {}, "color: _exists || size: _exists",
+                           {}, sort_fields, {0}).get();
+    ASSERT_EQ(3, results["found"].get<size_t>());
```
happy-san, line 4536: We should add
```c++
   std::vector<std::string> expected_ids = {"2", "1", "0"};
   for (size_t i = 0; i < results["hits"].size(); i++) {
       ASSERT_EQ(expected_ids[i], results["hits"][i]["document"]["id"].get<std::string>());
   }
```
to make sure only the expected ids match, in order.

test/collection_filtering_test.cpp
```diff
@@ -4258,4 +4258,445 @@ TEST_F(CollectionFilteringTest, DeepNestedObjectFieldsFiltering) {
...
+
+    ASSERT_TRUE(coll->add(R"({"id": "0", "title": "A", "rating": 4.5, "stock": 100, "in_stock": true, "points": 10})"_json.dump()).ok());
+    ASSERT_TRUE(coll->add(R"({"id": "1", "title": "B", "rating": 3.2, "in_stock": false, "points": 20})"_json.dump()).ok());
+    ASSERT_TRUE(coll->add(R"({"id": "2", "title": "C", "stock": 50, "points": 30})"_json.dump()).ok());
+    ASSERT_TRUE(coll->add(R"({"id": "3", "title": "D", "points": 40})"_json.dump()).ok());
+
+    // float exists
+    auto results = coll->search("*", {}, "rating: _exists", {}, sort_fields, {0}).get();
+    ASSERT_EQ(2, results["found"].get<size_t>());
+
```
happy-san, line 4573: [Same check](https://github.com/typesense/typesense/pull/2777/changes#r2816822665) in these tests as well.

test/collection_filtering_test.cpp
```diff
@@ -4258,4 +4258,445 @@ TEST_F(CollectionFilteringTest, DeepNestedObjectFieldsFiltering) {
...
+    // add optional field with optional_index
+    auto alter_payload = R"({
+        "fields": [
+            {"name": "color", "type": "string", "optional": true, "optional_index": true}
+        ]
+    })"_json;
+    ASSERT_TRUE(coll->alter(alter_payload).ok());
+
+    // all existing docs should be missing
```
happy-san: We should have a document that has a color field prior to the alter to make sure that it will get indexed in the optional_index after alter.

include/index.h
```diff
@@ -458,6 +458,10 @@ class Index {
     id_list_t* seq_ids;
     mutable std::shared_mutex seq_ids_mutex;
 
+    // optional field existence indexes (only for fields with optional_index=true)
+    spp::sparse_hash_map<std::string, id_list_t*> field_exists_index;
+    spp::sparse_hash_map<std::string, id_list_t*> field_missing_index;
```
happy-san, line 494: Instead of having separate `field_exists_index` and `field_missing_index`, we can have a single index similar to boolean index and let `0` and `1` represent missing and existing field respectively. We can also get away with only keeping track of the seq_ids where the field is missing and use similar logic to how we get `NOT_EQUALS` matches for other filters.

kishorenc: > We can also get away with only keeping track of the seq_ids where the field is missing and use similar logic to how we get NOT_EQUALS matches for other filters.

This would mean keeping just `field_missing_index` and use the `NOT_EQUALS` logic for the exists condition. @ozanarmagan I think we should try this approach.

ozanarmagan: I liked the idea! Great

src/filter_result_iterator.cpp
```diff
@@ -1146,6 +1146,64 @@ void filter_result_iterator_t::init(const bool& enable_lazy_evaluation, const bo
...
+                existence_list_ptr = list;
+                seq_id = all_seq_ids_iterator.id();
+                approx_filter_ids_length = list->num_ids();
+            } else {
+                is_filter_result_initialized = true;
+                validity = invalid;
+            }
+        } else {
+            // _exists: all docs minus the missing list
+            // Get the complement of missing ids to get the existing ids.
```
happy-san: @ozanarmagan We should also implement iterative logic in case of `enable_lazy_filter` [like we evaluate integer filterr](https://github.com/typesense/typesense/blob/d09678852b1dfa3d0c0b96b651f56d465e76dc00/src/filter_result_iterator.cpp#L1288). 

You can refer to [this test](https://github.com/typesense/typesense/blob/d09678852b1dfa3d0c0b96b651f56d465e76dc00/test/filter_test.cpp#L1328-L1355) for details. The crux of the iterative logic is to return the seq_ids in between of the actual matches of the iterator. So if the iterator matches 0, 2, 5, ... The matches for not equals will be 1, 3, 4 ...

alangmartini:
Any chance we can expand this for arrays, so we can filter for empty/non-empty arrays? Or should we create a separate issue for the next iteration?

src/filter_result_iterator.cpp
ozanarmagan: Ready to review again

ozanarmagan:
> Any chance we can expand this for arrays, so we can filter for empty/non-empty arrays? Or should we create a separate issue for the next iteration?

@alangmartini Could you create another issue for this? I will address that in another PR.

src/filter_result_iterator.cpp
```diff
@@ -1150,55 +1154,76 @@ void filter_result_iterator_t::init(const bool& enable_lazy_evaluation, const bo
         auto map_it = index->field_missing_index.find(a_filter.field_name);
 
         if(a_filter.apply_not_equals) {
```
happy-san, line 1328: Let's add a comment here explaining that `apply_not_equals` is the default/positive case in this code path since we store seq_ids for `!_exists` in the index.

happy-san (commented):
Rest looks good!

happy-san:
@kishorenc PR is ready for your review.

kishorenc:
A bit of manual + automated review but I have gone through each issue carefully and verified that proposed fixes are logical. I have attached a patch. @happy-san please review and confirm. 

[2777_review_patch.patch](https://github.com/user-attachments/files/26296899/2777_review_patch.patch)


## Issues found

1. `optional_index` metadata was updated from partial update payloads instead of the final merged document state. In mixed update batches this could incorrectly mark an unchanged optional field as missing, breaking `field: _exists` / `field: !_exists`.
2. `optional_index: true` was accepted on `index: false` fields. That schema produced no backing existence index, so `_exists` / `!_exists` could return incorrect results.
3. `_exists` parsing used substring matching, so ordinary string filters containing `_exists` (for example `title: pre_exists_post`) were misparsed as existence filters.
4. `field::field_from_json()` did not preserve `optional_index`, so JSON-to-field reconstruction silently dropped the flag.
5. Lazy `_exists` iterators could not be reset or materialized correctly. `compute_iterators()` and related paths treated the missing-id list as the final result set, so lazy `_exists` searches could return wrong hits.

## Fix summary

- Existence bookkeeping now uses each record's final document state (`new_doc` for updates, `doc` for inserts) and updates the missing-field index in one place.
- Schema validation now rejects `optional_index` unless the field is both `optional: true` and `index: true`.
- Existence parsing now only triggers on exact `_exists` / `!_exists` tokens, while normal string filters continue to work.
- `field::field_from_json()` now carries `optional_index` through correctly.
- Lazy `_exists` iteration now has dedicated reset and materialization logic, so `_exists` uses the complement of the missing-id list consistently in iterator and search paths.
- Added focused regression tests for mixed-batch updates, string literals containing `_exists`, invalid `optional_index` schemas, JSON field reconstruction, lazy iterator materialization, and lazy search hits.

happy-san:
@ozanarmagan I have listed the tests that will surface each issue:
1. Have a schema like,
```json
{
  "fields": [
    {
      "name": "field",
      "type": "string"
    },
    {
      "name": "optional_field",
      "type": "string",
      "optional": true,
      "optional_index": true
    }
  ]
}
```
Add a document like,
```json
{
  "field": "foo",
  "optional_field": "bar"
}
```
Check that `filter_by: optional_field: !_exists` should match no document.
Update the document with:
```json
{
  "field": "baz"
}
```
`filter_by: optional_field: !_exists` should still match no document.
2. A simple test that should fail when creating a field with `optional_index: true` and `index: false` should work.
3. Using the schema ,
```json
{
  "fields": [
    {
      "name": "field_exists",
      "type": "string"
    },
    {
      "name": "field",
      "type": "string"
    }
  ]
}
```
try passing a filter like `field_exists: foo` or `field: value_exists`.
4. A test like https://github.com/typesense/typesense/blob/v31/test/collection_join_test.cpp#L6013-L6021 will surface this issue. We must add this test when a new option is added in the field.
5. Add 
```c++
    iter_exists.reset();
    ASSERT_EQ(filter_result_iterator_t::valid, iter_exists.validity);

    for (uint32_t i = 0; i < validate_ids.size(); i++) {
        ASSERT_EQ(filter_result_iterator_t::valid, iter_exists.validity);
        ASSERT_EQ(expected[i], iter_exists.is_valid(validate_ids[i]));

        if (expected[i] == 1) {
            iter_exists.next();
        }
        ASSERT_EQ(seq_ids[i], iter_exists.seq_id);
    }
    ASSERT_EQ(filter_result_iterator_t::invalid, iter_exists.validity);
```
 [after this line](https://github.com/ozanarmagan/typesense/blob/21f293d1bc158fe79ca0548328ae0da5fe212fcf/test/collection_filtering_test.cpp#L4785)

kishorenc:
The file I have attached to my previous comment already has the patch that contains both the code and the tests for these issues.

src/filter_result_iterator.cpp
```diff
@@ -1310,6 +1315,87 @@ void filter_result_iterator_t::init(const bool& enable_lazy_evaluation, const bo
         return;
     }
 
+    if(!a_filter.comparators.empty() && a_filter.comparators[0] == MISSING) {
+        auto map_it = index->field_missing_index.find(a_filter.field_name);
+
+        if(!a_filter.apply_not_equals) {
```
happy-san: @ozanarmagan Instead of a double negative if condition, we should have `if (a_filter.apply_not_equals)` and exchange the contents of the if-else branches.

src/filter.cpp
```diff
@@ -517,6 +517,36 @@ Option<bool> toFilter(const std::string& expression,
     const field& _field = field_it.value();
     std::string&& raw_value = expression.substr(found_index + 1, std::string::npos);
     StringUtils::trim(raw_value);
+
+    // Handle missing filter: field: _missing  or  field: !_missing
+    std::string trimmed = raw_value;
+    StringUtils::trim(trimmed);
```
happy-san: `raw_value` is already trimmed at line 519.

src/filter.cpp
```diff
@@ -517,6 +517,36 @@ Option<bool> toFilter(const std::string& expression,
...
+                return Option<bool>(400, "Missing filter can only be applied to optional fields with `track_missing_values` enabled in the schema.");
+            }
+
+            filter_exp = {field_name, {}, {MISSING}};
+            filter_exp.apply_not_equals = is_negated;
+
+            return Option<bool>(true);
+        }
+
+        if(StringUtils::begins_with(missing_token, "_missing")) {
```
happy-san: Let's pull the literal `"_missing"` here and on line 534 into a `static const std::string`.
I don't think there is a need for this check. A filter like, `filter_by: field: _missing value to filter` does not break any of the existing or new functionality.

src/filter.cpp
```diff
@@ -517,6 +517,36 @@ Option<bool> toFilter(const std::string& expression,
...
+            return Option<bool>(400, "Invalid syntax for missing filter.");
+        }
+    }
```
happy-san, line 542: We should also add some parsing unit tests like: 
1. Setting `_missing` filter for a normal field.
2. Setting `field: _missing` for a field having `track_missing_value` enabled.
3. ``field: `_missing` `` filter should not be parsed as a missing value filter.
4. `field: _missing value` also should not be parsed as a missing value filter.

test/collection_manager_test.cpp
```diff
@@ -217,6 +221,7 @@ TEST_F(CollectionManagerTest, CollectionCreation) {
               "name":"location",
               "nested":false,
               "optional":true,
+              "track_missing_values":false,
```
happy-san: Let's add `"track_missing_values": true` at line 58 and also make the same change in the `expected_meta_json`.

test/collection_filtering_test.cpp
```diff
@@ -4258,4 +4258,793 @@ TEST_F(CollectionFilteringTest, DeepNestedObjectFieldsFiltering) {
...
+        ASSERT_EQ(expected[i] == -1 ? filter_result_iterator_t::invalid : filter_result_iterator_t::valid,
+                  iter_missing.validity);
+        ASSERT_EQ(expected[i], iter_missing.is_valid(validate_ids[i]));
+
+        if (expected[i] == 1) {
+            iter_missing.next();
+        }
+        ASSERT_EQ(seq_ids[i], iter_missing.seq_id);
+    }
+    ASSERT_EQ(filter_result_iterator_t::invalid, iter_missing.validity);
```
happy-san, line 4975: We are still missing any test that goes through `.reset()`. Let's add these lines here:
```c++
    iter_missing.reset();
    ASSERT_EQ(filter_result_iterator_t::valid, iter_missing.validity);

    for (uint32_t i = 0; i < validate_ids.size(); i++) {
        ASSERT_EQ(expected[i] == -1 ? filter_result_iterator_t::invalid : filter_result_iterator_t::valid,
                  iter_missing.validity);
        ASSERT_EQ(expected[i], iter_missing.is_valid(validate_ids[i]));

        if (expected[i] == 1) {
            iter_missing.next();
        }
        ASSERT_EQ(seq_ids[i], iter_missing.seq_id);
    }
    ASSERT_EQ(filter_result_iterator_t::invalid, iter_missing.validity);
```

src/filter_result_iterator.cpp
```diff
@@ -1310,6 +1315,87 @@ void filter_result_iterator_t::init(const bool& enable_lazy_evaluation, const bo
...
+            approx_filter_ids_length = num_ids - missing_count;
+
+            if (enable_lazy_evaluation && approx_filter_ids_length >= numeric_filter_ids_threshold) {
+                is_missing_filter = true;
+                missing_list_ptr = nullptr;
+                if(map_it != index->field_missing_index.end() && map_it->second != nullptr) {
+                    missing_list_ptr = map_it->second;
+                }
+                last_valid_id = index->seq_ids->last_id();
+                reset_missing_iterator();
```
happy-san: For lazy evaluation of `!_missing` filter, `is_not_equals_iterator` should be set to true.

src/filter_result_iterator.cpp
```diff
@@ -1034,6 +1034,11 @@ void filter_result_iterator_t::next() {
 
     const filter a_filter = filter_node->filter_exp;
 
+    if (is_missing_filter) {
```
happy-san, line 1041: We should move 
```c++
    if (is_not_equals_iterator) {
        return;
    }
```
from line 1060 to be earlier than the `if (is_missing_filter)` check.

src/filter_result_iterator.cpp
```diff
@@ -2415,6 +2501,24 @@ int filter_result_iterator_t::is_valid(uint32_t id, const bool& curation_timeout
...
+        seq_id = id + 1;
+
+        // _missing: match if id IS in the missing list
+        // !_missing: match if id is NOT in the missing list (field exists)
+        bool is_match = filter_node->filter_exp.apply_not_equals
+            ? (missing_list_ptr == nullptr || !missing_list_ptr->contains(id))
+            : (missing_list_ptr != nullptr && missing_list_ptr->contains(id));
+
+        return is_match ? 1 : 0;
+    }
```
happy-san: We should move this entire block into `skip_to(id)` method. The logic for `is_missing_filter` should look similar to the `else if (f.is_bool())` block in the `skip_to(id)` method.

include/filter_result_iterator.h
```diff
@@ -329,6 +329,24 @@ class filter_result_iterator_t {
 
     bool delete_filter_node = false;
 
+    bool is_missing_filter = false;
+    id_list_t* missing_list_ptr = nullptr;
+
```
happy-san, line 335: We should have a `id_list_t::iterator_t missing_values_iterator` for lazy evaluation of missing filter.

include/filter_result_iterator.h
```diff
@@ -329,6 +329,24 @@ class filter_result_iterator_t {
...
+    void compute_result_from_id_list(id_list_t* source);
+
+    /// Resets the iterator state for missing filters.
+    void reset_missing_iterator();
+
+    /// Advances the iterator state for missing filters.
+    void advance_missing_iterator();
+
+    /// Computes the full result for missing filters.
+    void compute_missing_result();
```
happy-san: Let's remove these methods. The logic for missing filter iterator will be similar to that of `id` filter iterator.

src/filter_result_iterator.cpp
```diff
@@ -2556,16 +2802,13 @@ void filter_result_iterator_t::reset(const bool& curation_timeout) {
...
-        all_seq_ids_iterator = index->seq_ids->new_iterator();
-        if (all_seq_ids_iterator.valid()) {
-            seq_id = all_seq_ids_iterator.id();
-            approx_filter_ids_length = index->seq_ids->num_ids();
-            validity = valid;
-        } else {
-            validity = invalid;
-        }
+    if (is_missing_filter) {
+        reset_missing_iterator();
```
happy-san: Only need to get new value for  `missing_values_iterator` from `missing_list_ptr.new_iterator();`.
In case of `!_missing` iterator, set `seq_id = 0`.

src/filter_result_iterator.cpp
```diff
@@ -1034,6 +1034,11 @@ void filter_result_iterator_t::next() {
 
     const filter a_filter = filter_node->filter_exp;
 
+    if (is_missing_filter) {
+        advance_missing_iterator();
```
happy-san: Only need to do `missing_values_iterator.next()` in this block, similar to `id` filter iterator.

src/filter_result_iterator.cpp
```diff
@@ -3026,29 +3269,13 @@ void filter_result_iterator_t::compute_iterators() {
...
-        }
-
-        is_filter_result_initialized = true;
-
-        if (validity != timed_out && filter_result.count == 0) {
-            validity = invalid;
-            return;
-        }
+    if (is_missing_filter) {
+        compute_missing_result();
```
happy-san: Only need to uncompress the missing_list_ptr and in case of `!_missing`, need to negate the results.

test/collection_filtering_test.cpp
```diff
@@ -4258,4 +4258,793 @@ TEST_F(CollectionFilteringTest, DeepNestedObjectFieldsFiltering) {
...
+                                               enable_lazy_evaluation);
+    ASSERT_TRUE(iter_missing.init_status().ok());
+    ASSERT_FALSE(iter_missing._get_is_filter_result_initialized());
+
+    // !_missing (complement): docs with color = {0, 1, 5}
+    // is_valid sets seq_id = id+1 for the probed doc; next() advances the complement iterator on matches.
+    // is_valid(0)=1, next()→1; is_valid(1)=1, next()→5; is_valid(2)=0; is_valid(3)=0; is_valid(4)=0;
+    // is_valid(5)=1, next()→invalid
+    std::vector<uint32_t> validate_ids = {0, 1, 2, 3, 4, 5, 6};
+    std::vector<uint32_t> seq_ids = {1, 5, 3, 4, 5, 6, 6};
```
happy-san: The `seq_ids` here should be `{1, 2, 3, 4, 5, 6, 6}` after the changes.

happy-san (commented):
@ozanarmagan I've left some comments for lazy evaluation path of missing filter. The logic is supposed to be similar to `id` field evaluation for `_missing` lazy filter and for `!_missing` lazy filter, the logic will be similar to `!=` lazy numeric filter. 

Let me know if I can clarify anything further.

kishorenc (approved):

PR #2781, opened 2026-02-18
don't transliterate stopword tokens

krunal1313:
## PR Checklist

### DIFF
--- a/src/tokenizer.cpp
+++ b/src/tokenizer.cpp
@@ -6,9 +6,10 @@
 
 Tokenizer::Tokenizer(const std::string& input, bool normalize, bool no_op, const std::string& locale,
                      const std::vector<char>& symbols_to_index,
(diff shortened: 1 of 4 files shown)


src/tokenizer.cpp
```diff
@@ -131,8 +132,11 @@ bool Tokenizer::next(std::string &token, size_t& token_index, size_t& start_inde
                     raw_text = icu::UnicodeString::fromUTF8(stemmed_word);
                 }
 
-                auto transliterator = TransliteratorPool::get_instance().acquire("Any-Latin;Latin-ASCII");
-                transliterator->transliterate(raw_text);
+                if (!is_stopword) { //dont transliterate stopword tokens
```
kishorenc: Instead of calling the parameter `is_stopword` it can be more generic, i.e. `do_transliterate` which can be true by default and is only disabled for stopword tokenization.

kishorenc (approved):

PR #2782, opened 2026-02-18
Optimize asymetric AND filters via sort_index probing

anthonybachour:
### Problem

When a highly selective equality filter is ANDed with a broad range filter on a `range_index` field, Typesense eagerly materializes all matching document IDs from the range index trie before intersecting. For large collections, this is extremely expensive — a filter like `price:<=5000` on a 276M-document collection allocates and populates an array of ~200M+ IDs, even when the other side of the AND matches only ~1,000 documents.

This makes common query patterns like `category:=X && price:<=Y` take 4-8 seconds on large collections, despite the final result set being tiny.

### Solution

Three changes to `filter_result_iterator`:

1. **Lazy range_index materialization** (`init()`): When `enable_lazy_evaluation` is true, defer the expensive trie search for `range_index` fields. Use `trie->size()` as an approximate count for optimization heuristics instead.

2. **Targeted lazy deferral** (constructor): For AND operators, force lazy evaluation on children that are leaf nodes with `range_index` fields, so the asymmetric probe optimization can bypass materialization entirely.

3. **Asymmetric AND probe** (`compute_iterators()`): When one side of an AND is 10x+ larger than the other and the small side has <500K results, compute the small side eagerly and probe each document against the large side via `sort_index` hash map lookups (O(1) per document). Falls back to `is_valid()` probing when sort_index isn't available, and to standard `and_filter_results()` when the optimization doesn't apply.

### Benchmark

Tested on a single-node collection with **276M documents** (ARM64, 128GB RAM). Each document has an int32 equality-filter field and an int32 range-indexed + sorted numeric field.

**Selective equality + range filter (the target query pattern):**

| Query | Stock v30.1 | Optimized | Speedup |
|---|---|---|---|
| `eq_field:=X && num_field:<=500` (99 matches) | 248ms | 0ms | >248x |
| `eq_field:=X && num_field:<=2000` (1,727 matches) | 4,469ms | 1ms | 4,469x |
| `eq_field:=X && num_field:<=5000` (2,679 matches) | 7,909ms | 1ms | 7,909x |
| `eq_field:=X && num_field:<=5000` (243 matches) | 7,924ms | 0ms | >7,924x |
| `eq_field:=X && num_field:[500..2000]` (1,639 matches) | ~4,500ms | 1ms | ~4,500x |

**Regression check — queries unaffected by optimization:**

| Query | Stock v30.1 | Optimized |
|---|---|---|
| `num_field:<=500` (standalone, 5.4M matches) | 376ms | 327ms |

Result counts are identical between stock and optimized builds across all queries.

### Files changed

- `include/filter.h` — Named constants `ASYMMETRIC_AND_RATIO_THRESHOLD` and `ASYMMETRIC_AND_MAX_PROBE_COUNT`
- `include/filter_result_iterator.h` — `range_index_deferred` flag
- `src/filter_result_iterator.cpp` — Lazy init, deferred child detection, asymmetric AND probe with sort_index

### DIFF
--- a/include/filter.h
+++ b/include/filter.h
@@ -14,6 +14,16 @@
 
 constexpr size_t DEFAULT_FILTER_BY_CANDIDATES = 4;
 
+/// Minimum ratio (large_side / small_side) required to trigger the asymmetric AND
+/// filter optimization, which probes the large side via sort_index instead of
+/// materializing it. Higher values are more conservative (fewer false triggers).
+constexpr uint32_t ASYMMETRIC_AND_RATIO_THRESHOLD = 10;
+
+/// Maximum number of documents on the small side for the asymmetric AND optimization.
+/// Beyond this, the per-document probe loop becomes too expensive relative to the
+/// standard sorted intersection.
+constexpr uint32_t ASYMMETRIC_AND_MAX_PROBE_COUNT = 500'000;
+
 enum NUM_COMPARATOR {
     LESS_THAN,
     LESS_THAN_EQUALS,
--- a/include/filter_result_iterator.h
+++ b/include/filter_result_iterator.h
@@ -320,6 +320,10 @@ private:
 
     bool delete_filter_node = false;
 
+    /// Set when init() deferred range_index materialization for lazy evaluation.
+    /// compute_iterators() will materialize via init(false, false) if needed.
+    bool range_index_deferred = false;
+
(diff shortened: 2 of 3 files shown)


anthonybachour:
Just FYI there's a but with filter_by not being applied properly I will update today with fixed version

PR #2793, opened 2026-02-24
fix: ensure forward-only iterator reads lower seq_id first in diversity similarity

alangmartini:
## Summary

MMR diversity reranking via curation sets silently fails when the selected document has a lower internal `seq_id` than the candidate document.

**Root cause:** `posting_list_t::iterator_t::skip_to()` is forward-only — it can only advance, never reverse. In `similarity_t::calculate()`, when `seq_id_j < seq_id_i`, the second `skip_to(seq_id_j)` stays at `seq_id_i`'s position because it cannot go backward. The subsequent ID check (`facet_it.id() != seq_id_j`) fails, and the function silently returns similarity = 0 for that facet field.

This defeats diversity reranking in any scenario where a previously selected document has a lower `seq_id` than the current candidate — most notably with `sort_by=...:asc`, where the first selected document (position 0) has the lowest `seq_id` and all subsequent candidates have higher IDs.

**Fix:** Use `std::min(seq_id_i, seq_id_j)` / `std::max(seq_id_i, seq_id_j)` to always read the lower `seq_id` first, ensuring the forward-only iterator always advances correctly. Jaccard and equality similarity are symmetric (`sim(A,B) == sim(B,A)`), so reading order does not affect correctness.

## Changes

- **`src/diversity.cpp`** — Read `min(seq_id_i, seq_id_j)` first and `max(...)` second in `similarity_t::calculate()`, guaranteeing the iterator always moves forward.
- **`test/collection_curation_test.cpp`** — Added `DiversityForwardOnlyIteratorBug` regression test that reproduces the bug using `sort_by=sort_order:asc` with `diversity_lambda=0`. Also updated expected assertions in the existing `DiversityOverride` test to reflect the correct diversity ordering now that similarity is computed correctly in both iterator directions.

## Test plan

- [ ] New `DiversityForwardOnlyIteratorBug` test passes (fails without the fix, passes with it)
- [ ] Existing `DiversityOverride` test passes with updated assertions
- [ ] Full `collection_curation_test` suite passes
- [ ] Verify no regressions in other diversity/curation tests

### DIFF
--- a/src/diversity.cpp
+++ b/src/diversity.cpp
@@ -104,8 +104,16 @@ Option<double> similarity_t::calculate(uint32_t seq_id_i, uint32_t seq_id_j, con
                 if (!facet_it.valid()) {
                     continue;
                 }
-                facet_it.skip_to(seq_id_i);
-                if (!facet_it.valid() || facet_it.id() != seq_id_i) {
+
+                // The posting list iterator is forward-only: skip_to() can only
+                // advance, never reverse. Read the lower seq_id first to ensure
+                // the second skip_to always moves forward. Jaccard/equality are
+                // symmetric so the order doesn't affect the result.
+                uint32_t first_id = std::min(seq_id_i, seq_id_j);
+                uint32_t second_id = std::max(seq_id_i, seq_id_j);
+
+                facet_it.skip_to(first_id);
+                if (!facet_it.valid() || facet_it.id() != first_id) {
                     continue;
                 }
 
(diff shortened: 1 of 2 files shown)


alangmartini:
Fixes https://github.com/typesense/typesense/issues/2791

cc: @kishorenc

kishorenc:
@happy-san Please review.

happy-san:
Looks good. It is the correct fix. You can also achieve the same with 
```c++
    if (seq_id_j < seq_id_i) {
        std::swap(seq_id_i, seq_id_j);
    }
```

happy-san:
Sidenote: [I wasn't sure about having a cache for the similarity score I had in the first implementation.](https://github.com/typesense/typesense/pull/2572/changes#diff-563711aec2e1db6e7399cfaf31fc8c6e9e465e08b405ebb79db5bb194e44267eR95-R103) I checked the cache hit ratio was about 50% for the example that I tried it with so I decided to drop it. @alangmartini can you take a look so we can be sure about this as well?

alangmartini:
@happy-san yeah, it makes sense to me 🙇! The cache doesn't make more sense in the current implementation, since max_similarities[i] tracks each candidate maximum and there's nothing to cache.

kishorenc (approved):

PR #2801, opened 2026-03-01
fix: validate int32 range on document write

### LINKED ISSUE #2798: int32 fields accept values exceeding INT32_MAX on write but reject them on search/filter
### Description

When a field is defined as `int32`, Typesense allows indexing documents with values exceeding `INT32_MAX` (2,147,483,647) without returning a validation error. However, attempting to search or filter on those out-of-range values returns a `400 Bad Request` error.

This is a validation inconsistency. If the value is out of range, it should be rejected at **write time**, not silently accepted and then rejected at query time.

### Steps to reproduce

Save and run the following script (requires Docker, curl, and bash):

```bash
#!/bin/bash
# Reproducer: int32 fields accept values > INT32_MAX on write but reject on search
# Typesense Version: 28.0

set -e

TYPESENSE_API_KEY=xyz
PORT=8108
TYPESENSE_HOST=http://localhost:${PORT}
CONTAINER_NAME=typesense-issue-int32-overflow
DATA_DIR=$(pwd)/typesense-data-${CONTAINER_NAME}

cleanup() {
  echo ""
  echo "=== Cleanup ==="
  docker stop ${CONTAINER_NAME} 2>/dev/null || true
  docker rm ${CONTAINER_NAME} 2>/dev/null || true
  rm -rf ${DATA_DIR}
}
trap cleanup EXIT

docker stop ${CONTAINER_NAME} 2>/dev/null || true
docker rm ${CONTAINER_NAME} 2>/dev/null || true
mkdir -p ${DATA_DIR}

docker run -d \
  --name ${CONTAINER_NAME} \
  -p ${PORT}:8108 \
  -v ${DATA_DIR}:/data \
  typesense/typesense:30.1 \
  --data-dir /data \
  --api-key=${TYPESENSE_API_KEY} \
  --enable-cors > /dev/null

echo "Waiting for Typesense..."
until curl -s "${TYPESENSE_HOST}/health" 2>/dev/null | grep -q '"ok":true'; do sleep 1; done
echo "Typesense is ready"

### Create collection with int32 field
echo ""
echo "=== Creating collection with int32 'timestamp' field ==="
curl -s "${TYPESENSE_HOST}/collections" \
  -X POST \
  -d '{
    "name": "events",
    "fields": [
      {"name": "timestamp", "type": "int32"}
    ]
  }' > /dev/null
sleep 2

INT32_MAX=2147483647
OVER_INT32=1740000000000  # ~2025 timestamp in milliseconds

echo ""
echo "=== Indexing document with valid int32 value (${INT32_MAX}) ==="
curl -s "${TYPESENSE_HOST}/collections/events/documents" \
  -X POST \
  -d "{\"id\": \"1\", \"title\": \"Normal event\", \"timestamp\": ${INT32_MAX}}"
echo ""

echo ""
echo "=== Indexing document with value EXCEEDING int32 (${OVER_INT32}) ==="
echo "BUG: This should fail validation but succeeds:"
curl -s "${TYPESENSE_HOST}/collections/events/documents" \
  -X POST \
  -d "{\"id\": \"2\", \"title\": \"Overflowed event\", \"timestamp\": ${OVER_INT32}}"
echo ""

echo ""
echo "=== Searching with filter on valid int32 value — should work ==="
curl -s "${TYPESENSE_HOST}/collections/events/documents/search" \
  -G \
  -d "q=*" \
  -d "query_by=title" \
  -d "filter_by=timestamp:=${INT32_MAX}"
echo ""

set +e
echo ""
echo "=== Searching with filter on overflowed value — BUG: returns 400 ==="
curl -s "${TYPESENSE_HOST}/collections/events/documents/search" \
  -G \
  -d "q=*" \
  -d "query_by=title" \
  -d "filter_by=timestamp:=${OVER_INT32}"
echo ""

echo ""
echo "=== Filtering with range on overflowed value — BUG: returns found:0 ==="
curl -s "${TYPESENSE_HOST}/collections/events/documents/search" \
  -G \
  -d "q=*" \
  -d "query_by=title" \
  -d "filter_by=timestamp:>${INT32_MAX}"
echo ""

echo ""
echo "=== Export (no filter) — works, overflowed value is retrievable ==="
curl -s "${TYPESENSE_HOST}/collections/events/documents/export" \
  -H "X-TYPESENSE-API-KEY: ${TYPESENSE_API_KEY}"
echo ""

echo ""
echo "=== Export with filter — BUG: same 400 error ==="
curl -s "${TYPESENSE_HOST}/collections/events/documents/export?filter_by=timestamp:=${OVER_INT32}" \
  -H "X-TYPESENSE-API-KEY: ${TYPESENSE_API_KEY}"
echo ""
set -e

echo ""
echo "=== Sort by timestamp — shows overflowed value stored as-is ==="
curl -s "${TYPESENSE_HOST}/collections/events/documents/search" \
  -G \
  -d "q=*" \
  -d "query_by=title" \
  -d "sort_by=timestamp:desc"
echo ""

echo ""
echo "============================================================"
echo "SUMMARY"
echo "============================================================"
echo "1. int32 field accepted value ${OVER_INT32} on write (no error)"
echo "2. Filtering on that value returns 400 on search and export"
echo "3. Range filter timestamp:>${INT32_MAX} returns found:0 (doc invisible)"
echo "4. Unfiltered export retrieves the document correctly"
echo "5. Expected: write should reject values outside int32 range"
echo "   int32 range: -2147483648 to 2147483647"
echo "============================================================"
```

**Output (key lines):**

```
=== Indexing document with value EXCEEDING int32 (1740000000000) ===
BUG: This should fail validation but succeeds:
{"id":"2","timestamp":1740000000000,"title":"Overflowed event"}

=== Searching with filter on overflowed value — BUG: returns 400 ===
{"message": "Error with filter field `timestamp`: `1740000000000` exceeds the range of an int32."}

=== Filtering with range on overflowed value — BUG: returns found:0 ===
{"facet_counts":[],"found":0,"hits":[],...}

=== Export (no filter) — works, overflowed value is retrievable ===
{"id":"1","timestamp":2147483647,"title":"Normal event"}
{"id":"2","timestamp":1740000000000,"title":"Overflowed event"}
```

### Expected behavior

Indexing a document with a value outside the int32 range (`-2,147,483,648` to `2,147,483,647`) should return a validation error at write time, e.g.:

```json
{"message": "Field `timestamp` must be an int32, but the value `1740000000000` exceeds the range."}
```

### Actual behavior

| Operation | Result |
|---|---|
| Index document with `timestamp: 1740000000000` | Silently accepted |
| Filter `timestamp:=1740000000000` | `400: exceeds the range of an int32` |
| Filter `timestamp:>2147483647` | Returns `found: 0` (document is invisible to range filters) |
| Sort by `timestamp:desc` | Returns the document with `1740000000000` stored as-is |
| Export (no filter) | Returns the document with `1740000000000` correctly |
| Export with `filter_by=timestamp:=1740000000000` | Same 400 error |

The value is stored in the index and retrievable via unfiltered search/export, but it becomes **unfilterable** — the user cannot even query to find and clean up the bad documents without exporting the entire collection.

### Metadata

- **Typesense version:** 30.1 (likely affects all versions with int32 field type)
- **Use case:** Customer stored millisecond timestamps in an int32 field. Writes succeeded, but they cannot filter or search on those values.

alangmartini:
## Summary

Fixes #2798

**Bug:** int32 fields silently accept values exceeding `INT32_MAX` (e.g., millisecond timestamps like `1740000000000`) on document write, but reject them with HTTP 400 on search/filter. This causes a confusing experience where documents are indexed successfully but become unfilterable.

**Root cause:** In `validator.cpp`, when a JSON value is already a `number_integer`, the validator skips `coerce_int32_t` (which contains the range check) and accepts the value as-is. The range validation only fires when the value needs type coercion (e.g., from string or float).

**Fix:**
- Added explicit range validation in the `else` branch of the `is_number_integer()` check, for both scalar `INT32` and `INT32_ARRAY` fields. Values outside `[INT32_MIN, INT32_MAX]` are now rejected at write time with a clear 400 error, or silently dropped when `dirty_values` is `DROP`/`COERCE_OR_DROP` and the field is optional.
- Fixed the existing `coerce_int32_t` function to also check `INT32_MIN` bounds (previously only checked `INT32_MAX`).

## Test plan

- [x] **Scalar overflow test** (`Int32FieldRejectsOverflowOnWrite`): verifies valid boundary values are accepted, and values exceeding `INT32_MAX` or below `INT32_MIN` are rejected with 400
- [x] **Array overflow test** (`Int32ArrayFieldRejectsOverflowOnWrite`): verifies array elements exceeding int32 range are rejected
- [x] **dirty_values DROP test** (`Int32OverflowWithDirtyValuesDropOptional`): verifies optional fields with `DROP` silently discard overflowed values, while non-optional fields still reject
- [x] Full test suite passes (665.8s, all existing tests green)

### DIFF
--- a/include/collection.h
+++ b/include/collection.h
@@ -5,6 +5,7 @@
 #include <string>
 #include <unordered_map>
 #include <thread>
+#include <memory>
 #include <mutex>
 #include <condition_variable>
 #include <shared_mutex>
@@ -27,17 +28,34 @@ struct doc_seq_id_t {
     bool is_new;
 };
 
+struct highlight_query_token_t {
+    bool is_prefix = false;
+    uint32_t root_len = 0;
+    uint32_t num_typos = 0;
+
+    highlight_query_token_t() = default;
+
+    highlight_query_token_t(uint32_t root_len, uint32_t num_typos, bool is_prefix):
+            is_prefix(is_prefix), root_len(root_len), num_typos(num_typos) {
+    }
+};
+
 struct highlight_field_t {
     std::string name;
     bool fully_highlighted;
     bool infix;
     bool is_string;
-    tsl::htrie_map<char, token_leaf> qtoken_leaves;
+    tsl::htrie_map<char, highlight_query_token_t> qtoken_leaves;
 
     highlight_field_t(const std::string& name, bool fully_highlighted, bool infix, bool is_string):
             name(name), fully_highlighted(fully_highlighted), infix(infix), is_string(is_string) {
 
     }
+};
+
+struct highlight_field_snapshot_t {
+    tsl::htrie_map<char, highlight_query_token_t> qtoken_leaves;
+    std::vector<std::unique_ptr<posting_list_t>> posting_lists;
 };
 
 struct union_global_params_t {
@@ -477,7 +495,7 @@ private:
                                       const std::vector<char>& symbols_to_index, const std::vector<char>& token_separators,
                                       highlight_t& highlight, StringUtils& string_utils, const bool& use_word_tokenizer,
                                       const size_t& highlight_affix_num_tokens,
-                                      const tsl::htrie_map<char, token_leaf>& qtoken_leaves, const int& last_valid_offset_index,
+                                      const tsl::htrie_map<char, highlight_query_token_t>& qtoken_leaves, const int& last_valid_offset_index,
                                       const size_t& prefix_token_num_chars, const bool& highlight_fully,
                                       const size_t& snippet_threshold, const bool& is_infix_search,
                                       const std::vector<std::string>& raw_query_tokens, const size_t& last_valid_offset,
@@ -488,7 +506,7 @@ private:
     static void highlight_result(const bool& enable_nested_fields, const std::vector<char>& symbols_to_index,const std::vector<char>& token_separators,
                                  const std::string& raw_query, const field& search_field,
                                  const size_t& search_field_index,
-                                 const tsl::htrie_map<char, token_leaf>& qtoken_leaves,
+                                 const highlight_field_snapshot_t& highlight_snapshot,
                                  const KV* field_order_kv, const nlohmann::json& document,
                                  nlohmann::json& highlight_doc,
                                  StringUtils& string_utils,
@@ -511,7 +529,9 @@ private:
                                 const size_t& highlight_affix_num_tokens, const string& highlight_start_tag,
                                 const string& highlight_end_tag, const std::vector<std::string>& highlight_field_names,
                                 const std::vector<std::string>& highlight_full_field_names,
-                                const std::vector<highlight_field_t>& highlight_items, const uint8_t* index_symbols,
+                                const std::vector<highlight_field_t>& highlight_items,
+                                const std::vector<highlight_field_snapshot_t>& highlight_snapshots,
+                                const uint8_t* index_symbols,
                                 const KV* field_order_kv, const nlohmann::json& document, nlohmann::json& highlight_res,
                                 nlohmann::json& wrapper_doc,
                                 const std::vector<std::vector<std::string>>& q_phrases = {});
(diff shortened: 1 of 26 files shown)


alangmartini:
Closing in favor of #2803, which centralizes the int32 bounds validation into shared helpers and explicitly handles unsigned JSON integers.

PR #2802, opened 2026-03-01
fix: shared int32 bounds validation on write (alt to #2801)

alangmartini:
Alternative to #2801 for issue #2798.

This implementation takes a unified validation path for `int32`:
- Coerce type only when needed.
- Always run shared int32 bounds validation afterward for both scalar and array fields.
- Handle both signed and unsigned JSON integer representations explicitly.

Why this variant:
- Avoids special-case range checks only in `is_number_integer()` bypass branches.
- Keeps range enforcement as a single reusable step instead of duplicated branch logic.

Behavior covered:
- Reject values outside `[-2147483648, 2147483647]` on write.
- Keep optional-field + `DIRTY_VALUES::DROP` / `COERCE_OR_DROP` semantics (drop overflowed value).

Tests added:
- `Int32FieldRejectsOverflowOnWrite`
- `Int32ArrayFieldRejectsOverflowOnWrite`
- `Int32OverflowWithDirtyValuesDropOptional`

Note: local build/test execution was not completed in this Windows worktree due repo build-toolchain constraints (`libfor` dependency step expects Unix `make/cc`).

### DIFF
--- a/src/validator.cpp
+++ b/src/validator.cpp
@@ -1,6 +1,66 @@
 #include "validator.h"
 #include "field.h"
 
+namespace {
+bool should_drop_field(const field& a_field, const DIRTY_VALUES& dirty_values) {
+    return a_field.optional &&
+           (dirty_values == DIRTY_VALUES::DROP || dirty_values == DIRTY_VALUES::COERCE_OR_DROP);
+}
+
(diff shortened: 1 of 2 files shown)


alangmartini:
Superseded by #2803, which targets v30 for an apples-to-apples comparison with #2801.

PR #2813, opened 2026-03-04
feat: Infix filtering support #2685

### LINKED ISSUE #2685: [Feature Request] Infix filtering support
**Feature description**
It would be useful to be able to use infix within filter, so that we can match not only in an "exact" or "prefix" mode, but also a "contains" mode.

To be consistent with current prefix syntax, an infix filter may take the form: 

`filter_by: "author:*Jack*"`

**Use Case**
We can already search within words in our queries using the infix feature, provided we have activated the additional structures in our collection and we provide the right parameters at search time.

We however can't search within words in filtering, which is a limitation to our refinement of search results.

**Describe alternatives you've considered**
Possibly using multisearch and intersect results outside of typesense as we can only union as of now.

**Additional context**
This was already requested in https://github.com/typesense/typesense/issues/696, which did lead to the implementation of prefix support in filtering, but somehow infix was de-scoped from the original request.

mdhaduk:
---                                                                                                                                                              
  feat: infix filtering support (field:*value*) #2685
                                                                                                                                                                   
  Closes #2685                                    

  ---
  Typesense supports exact (field:value) and prefix (field:value*) filtering on string fields, but had no way to filter by substring , users couldn't express "give me all docs where this field _contains_ this value."

  This PR adds field:*value* syntax to filter_by, enabling infix/contains filtering on string and string array fields that have `infix: true` in their schema.

  Changes:

  - Detection (filter_result_iterator.cpp): *value* is now detected as an infix pattern before the existing prefix (value*) check, avoiding misidentification.
  - Validation: Returns a 400 error if the field does not have infix: true... the infix index is only built when explicitly enabled.
  - Lookup: Reuses the existing Index::search_infix() function (already used for query-time infix search). For multi-token filter values (e.g. *Chris P*), results
  are intersected (AND). Multiple filter values in a list (e.g. [*foo*, *bar*]) are OR'd together.
  - Result assembly: Results are stored as a flat filter_result_t (sorted uint32_t[]) -> the same pattern used by numeric and geo filters. Mixed infix + exact/prefix in a single OR list is handled by flattening posting lists and OR'ing with infix results before returning.

  Tesing (CollectionFilteringTest.InfixFilterOnTextFields)

  - Basic single-token infix on string[] field (*ris* matches "Chris Evans", "Chris Parnell", "Chris Pine")
  - Infix on string field
  - No matches case
  - OR of multiple infix values ([*ris*, *art*])
  - Error on field without infix: true (400 response)
  - Mixed infix + exact in OR ([*ris*, Martin])
  - Mixed infix + prefix in OR ([*ris*, Ma*])
  - Infix combined with AND on another field (cast: *ris* && points: >60)

### DIFF
--- a/test/collection_filtering_test.cpp
+++ b/test/collection_filtering_test.cpp
@@ -3397,7 +3397,8 @@ TEST_F(CollectionFilteringTest, InfixFilterOnTextFields) {
     auto res_op = coll->search("*", {}, "name_no_infix: *foo*", {}, {}, {0}, 10, 1, FREQUENCY, {false});
     ASSERT_FALSE(res_op.ok());
     ASSERT_EQ(400, res_op.code());
-    ASSERT_NE(std::string::npos, res_op.error().find("infix"));
+    ASSERT_EQ("Error with filter field `name_no_infix`: Infix filtering requires the field to have "
+              "`infix: true` in the schema.", res_op.error());
 
     // Test 6: Mixed infix + exact in OR - [*ris*, Martin]
     results = coll->search("*", {}, "cast: [*ris*, Martin]", {}, {}, {0}, 10, 1, FREQUENCY, {false}).get();
--- a/test/collection_filtering_test.cpp
+++ b/test/collection_filtering_test.cpp
@@ -3437,6 +3437,52 @@ TEST_F(CollectionFilteringTest, InfixFilterOnTextFields) {
     results = coll->search("*", {}, "title: *unt*", {}, {}, {0}, 10, 1, FREQUENCY, {false}).get();
     ASSERT_EQ(1, results["hits"].size());
     ASSERT_EQ("Good Will Hunting", results["hits"][0]["document"]["title"]);
+
+    // Test 10: enable_lazy_filter=true + OR of infix values, cast:[*an*,*in*] with q:"will".
+    {
+        std::map<std::string, std::string> req_params = {
+            {"collection", "InfixFilterColl"},
+            {"q", "will"},
+            {"query_by", "title"},
+            {"filter_by", "cast: [*an*, *in*]"},
+            {"enable_lazy_filter", "true"}
+        };
+        nlohmann::json embedded_params;
+        std::string json_res;
+        auto now_ts = std::chrono::duration_cast<std::chrono::microseconds>(
+            std::chrono::system_clock::now().time_since_epoch()).count();
+        auto search_op = collectionManager.do_search(req_params, embedded_params, json_res, now_ts);
+        ASSERT_TRUE(search_op.ok());
+        auto res_obj = nlohmann::json::parse(json_res);
+        ASSERT_EQ(1, res_obj["found"].get<size_t>());
+        ASSERT_EQ("2", res_obj["hits"][0]["document"]["id"].get<std::string>());
+    }
+
+    // Test 11: enable_lazy_filter=true + negated OR of infix values, cast:![*an*,*in*] with q:"will".
+    // NOT({0,2,4,5}) = {1,3}. "will" in title: {2,3}. Result = {3}.
+    {
+        std::map<std::string, std::string> req_params = {
+            {"collection", "InfixFilterColl"},
+            {"q", "will"},
+            {"query_by", "title"},
+            {"filter_by", "cast: ! [*an*, *in*]"},
+            {"enable_lazy_filter", "true"}
+        };
+        nlohmann::json embedded_params;
+        std::string json_res;
+        auto now_ts = std::chrono::duration_cast<std::chrono::microseconds>(
+            std::chrono::system_clock::now().time_since_epoch()).count();
+        auto search_op = collectionManager.do_search(req_params, embedded_params, json_res, now_ts);
+        ASSERT_TRUE(search_op.ok());
+        auto res_obj = nlohmann::json::parse(json_res);
+        ASSERT_EQ(1, res_obj["found"].get<size_t>());
+        ASSERT_EQ("3", res_obj["hits"][0]["document"]["id"].get<std::string>());
+    }
+
+    // Test 12: Exact-match operator (:=) with infix-style value — cast:= *in*.
+    // No exact matches to "in" in our cast so should return 0
+    results = coll->search("*", {}, "cast:= *in*", {}, {}, {0}, 10, 1, FREQUENCY, {false}).get();
+    ASSERT_EQ(0, results["hits"].size());
 }
 
 TEST_F(CollectionFilteringTest, ExactFilterOnLongField) {
--- a/test/collection_filtering_test.cpp
+++ b/test/collection_filtering_test.cpp
@@ -3330,6 +3330,114 @@ TEST_F(CollectionFilteringTest, PrefixFilterOnTextFields) {
     ASSERT_EQ("Steve Reiley", res_obj["hits"][3]["document"]["names"][0]);
 }
 
+TEST_F(CollectionFilteringTest, InfixFilterOnTextFields) {
+    // Create collection with infix: true on string fields
+    auto schema_json =
+            R"({
+                "name": "InfixFilterColl",
+                "fields": [
+                    {"name": "title", "type": "string", "infix": true},
+                    {"name": "cast", "type": "string[]", "infix": true},
+                    {"name": "name_no_infix", "type": "string"},
+                    {"name": "points", "type": "int32"}
+                ],
+                "default_sorting_field": "points"
+            })"_json;
+
+    auto collection_create_op = collectionManager.create_collection(schema_json);
+    ASSERT_TRUE(collection_create_op.ok());
+    auto coll = collection_create_op.get();
+
+    std::vector<nlohmann::json> documents = {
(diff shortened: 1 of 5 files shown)


tharropoulos:
CC: @kishorenc

tharropoulos:
Is this ready for review?

mdhaduk:
New contributor, so I just wanted a review of this draft before submitting as PR in case of any obvious pitfalls

Let me know if I should just submit as PR, I think repo does not run clean build on issue of draft PR

tharropoulos:
@happy-san Could you check this out?

mdhaduk:
Went ahead and just opened it since some time has passed!

happy-san:
Hey @mdhaduk, The PR looks good! It is missing implementation for handling lazy evaluation of filter_by though. Let me know if you want to take a stab at that. What needs to be done:
* Instead of collecting the seq_ids https://github.com/typesense/typesense/blob/v31/src/index.cpp#L3420-L3422 only need to collect the posting lists.
* Using the existing [advance_string_filter_token_iterators](https://github.com/typesense/typesense/blob/v31/src/filter_result_iterator.cpp#L679) and [get_string_filter_next_match](https://github.com/typesense/typesense/blob/v31/src/filter_result_iterator.cpp#L695) should work otherwise the logic to be implemented will be similar.
* Need to implement test cases similar to https://github.com/typesense/typesense/blob/v31/test/filter_test.cpp to verify everything.

mdhaduk:
Hey @happy-san, thanks for taking a look! I’m happy to finish out the missing implementation for lazy evaluation of `filter_by`

mdhaduk:
Hi @happy-san, implemented the lazy evaluation as requested. Here's a summary of what changed:                                                                                                                                                              
                                                                                                                                                                                                                                                                 
  Lazy evaluation path (filter_result_iterator.cpp)                                                                                                                                                                                                              
  - Replaced the eager search_infix() call (which materialised all doc IDs into a flat array) with search_infix_leaves(), which returns art_leaf* pointers without merging doc IDs.                                                                              
  - Each matching vocab token becomes one outer entry in posting_list_iterators, ORed lazily by the existing get_string_filter_next_match() via the CONTAINS comparator branch — no position verification is needed, which is correct for infix.                 
  - Removed the has_infix_results finalization block; the normal string_filter_ids_threshold check now applies uniformly to infix.                                                                                                                               

  search_infix_leaves() (index.h, index.cpp)
  - Extracted the scan logic from search_infix() into a new function that stops before merging doc IDs. search_infix() now delegates to it (no behaviour change for query-time infix search).

  Tests (test/filter_test.cpp)
  - Added FilterTest.InfixLazyEvaluation with 5 cases: basic lazy iteration, no-match, OR of two infix values, infix AND numeric, and a multi-token error case.

 assumptions made: 
- multi-token infix (cast:*foo bar*) was removed and replaced with a 400 error. The infix index stores individual word tokens only, so substring search across word boundaries is architecturally impossible... the old path was producing semantically wrong AND results with no adjacency guarantee and had no test coverage. Users can write cast:*foo* && cast:*bar* as separate conditions instead.

test/collection_filtering_test.cpp
```diff
@@ -3330,6 +3330,114 @@ TEST_F(CollectionFilteringTest, PrefixFilterOnTextFields) {
...
+    ASSERT_TRUE(result_ids.count("0"));
+    ASSERT_TRUE(result_ids.count("1"));
+    ASSERT_TRUE(result_ids.count("2"));
+    ASSERT_TRUE(result_ids.count("5"));
+
+    // Test 5: Error when field lacks infix: true
+    auto res_op = coll->search("*", {}, "name_no_infix: *foo*", {}, {}, {0}, 10, 1, FREQUENCY, {false});
+    ASSERT_FALSE(res_op.ok());
+    ASSERT_EQ(400, res_op.code());
+    ASSERT_NE(std::string::npos, res_op.error().find("infix"));
```
happy-san: We should match the error message verbatim here.

test/filter_test.cpp
```diff
@@ -2865,3 +2865,138 @@ TEST_F(FilterTest, FilterReferences) {
...
+    // boundaries is structurally impossible. Users should write cast:*ris* && cast:*pin* instead.
+    filter_op = filter::parse_filter_query("cast: *ris pin*", coll->get_schema(), store,
+                                           doc_id_prefix, filter_tree_root);
+    ASSERT_TRUE(filter_op.ok());  // parse succeeds — parser stores raw value, unaware of token count
+
+    auto iter_multi_token = filter_result_iterator_t(coll->get_name(), coll->_get_index(),
+                                                     filter_tree_root, enable_lazy_evaluation);
+    ASSERT_FALSE(iter_multi_token.init_status().ok());
+    ASSERT_EQ(400, iter_multi_token.init_status().code());
+    ASSERT_NE(std::string::npos, iter_multi_token.init_status().error().find("single token"));
```
happy-san: Same here, we should match the error message verbatim.

test/filter_test.cpp
```diff
@@ -2865,3 +2865,138 @@ TEST_F(FilterTest, FilterReferences) {
...
+    }
+    ASSERT_EQ(filter_result_iterator_t::invalid, iter_infix_and_numeric.validity);
+
+    delete filter_tree_root;
+    filter_tree_root = nullptr;
+
+    // Test 5: Multi-token infix must return 400 — *ris pin* tokenizes to ["ris", "pin"].
+    // The infix index stores individual word tokens only; substring search across word
+    // boundaries is structurally impossible. Users should write cast:*ris* && cast:*pin* instead.
+    filter_op = filter::parse_filter_query("cast: *ris pin*", coll->get_schema(), store,
```
happy-san, line 3012: Let's add a test case having `cast: *ris* && cast: *in*` which should match doc `id: 5`.

mdhaduk:
Just pushed the small updates regarding error messages + `cast: *ris* && cast: *in*` test case!

test/collection_filtering_test.cpp
```diff
@@ -3330,6 +3330,115 @@ TEST_F(CollectionFilteringTest, PrefixFilterOnTextFields) {
...
+    auto coll = collection_create_op.get();
+
+    std::vector<nlohmann::json> documents = {
+        R"({"title": "Captain America", "cast": ["Chris Evans", "Scarlett Johansson"], "name_no_infix": "foo", "points": 78})"_json,
+        R"({"title": "Anchorman", "cast": ["Chris Parnell", "Josh Lawson"], "name_no_infix": "bar", "points": 63})"_json,
+        R"({"title": "There Will Be Blood", "cast": ["Martin Stringer", "Jacob Stringer"], "name_no_infix": "baz", "points": 81})"_json,
+        R"({"title": "Good Will Hunting", "cast": ["Matt Damon", "Ben Affleck"], "name_no_infix": "qux", "points": 83})"_json,
+        R"({"title": "Percy Jackson", "cast": ["Logan Lerman", "Alexandra Daddario"], "name_no_infix": "quux", "points": 59})"_json,
+        R"({"title": "Quantum Quest", "cast": ["Chris Pine"], "name_no_infix": "corge", "points": 52})"_json,
+    };
```
happy-san, line 3358: Let's add these final tests as well
```json
{
  "q": "will",
  "query_by": "title",
  "filter_by": "cast: [*an*, *in*]",
  "enable_lazy_filter": true
}
```
which should match document `id: 2`.

```json
{
  "q": "will",
  "query_by": "title",
  "filter_by": "cast: ! [*an*, *in*]",
  "enable_lazy_filter": true
}
```
which should match document `id: 3`.

You can call search like this as well: https://github.com/typesense/typesense/blob/v31/test/collection_filtering_test.cpp#L181-L198

happy-san: We also haven't tested what happens when a filter like `cast:= *in*` is sent. `:=` is the exact match filter operator.

mdhaduk: That's a good point, will add this as well!

mdhaduk: > You can call search like this as well: https://github.com/typesense/typesense/blob/v31/test/collection_filtering_test.cpp#L181-L198

Ah I see because coll->search doesn't expose `enable_lazy_filter`...

mdhaduk:
Updated `collection_filtering_test` with those three additional tests as outlined

test/collection_filtering_test.cpp
happy-san: It does, but you would've ended up initializing way too much stuff to get to it https://github.com/typesense/typesense/blob/v31/include/collection.h#L1084

happy-san:
@kishorenc PR looks good to me. Ready for your review.

test/collection_filtering_test.cpp
mdhaduk: > It does, but you would've ended up initializing way too much stuff to get to it https://github.com/typesense/typesense/blob/v31/include/collection.h#L1084

 I see I missed it, that makes sense

mdhaduk:
Any update on this?

kishorenc (approved):

Issue #2816, opened 2026-03-05
Hybrid search ignores alpha and returns 0 results when phrase query [`""` exact match] has no keyword matches


onceLearner:
Here's the updated bug report:

---

## Bug Description

When performing a hybrid search (keyword + vector) with a **phrase query** (double-quoted `q`), if the phrase matches zero documents on the keyword side, the entire result set comes back empty — even though the vector search component would have returned relevant results on its own.

This effectively makes the `alpha` weight parameter meaningless for phrase queries, since the vector branch is never executed.

A common real-world scenario: a user in a **conversation/RAG search** pastes a snippet of their code into the search box, something like `"const result = await client.get()"`. The double quotes in their input turn this into a phrase query. The exact phrase likely won't exist in any document, so keyword search returns nothing — and because of this bug, the vector/semantic search that *would* have found relevant documentation or similar code examples is silently skipped. The LLM backing the conversation gets zero context and can't help.

## Reproduction Steps

```bash
# 1. Create collection with an embedding field
curl -X POST "http://localhost:8108/collections" \
  -H "Content-Type: application/json" \
  -H "X-TYPESENSE-API-KEY: ${TYPESENSE_API_KEY}" \
  -d '{
    "name": "docs",
    "fields": [
      { "name": "title", "type": "string" },
      { "name": "content", "type": "string" },
      { "name": "language", "type": "string", "facet": true },
      {
        "name": "embedding",
        "type": "float[]",
        "embed": {
          "from": ["title", "content"],
          "model_config": {
            "model_name": "ts/all-MiniLM-L12-v2"
          }
        }
      }
    ]
  }'


# 2. Hybrid search with a phrase query that has a typo — returns 0 hits
curl -G "http://localhost:8108/multi_search" \
  -H "X-TYPESENSE-API-KEY: ${TYPESENSE_API_KEY}" \
  --data-urlencode 'searches=[{
    "collection": "docs",
    "q": "\"data dod\"",
    "query_by": "title,content,embedding",
    "vector_query": "embedding:([], alpha: 0.5)",
    "filter_by": "language:en",
    "prefix": "false",
    "per_page": "16",
    "exclude_fields": "embedding"
  }]'

# 3. For comparison — pure vector search works fine and returns the document
curl -G "http://localhost:8108/multi_search" \
  -H "X-TYPESENSE-API-KEY: ${TYPESENSE_API_KEY}" \
  --data-urlencode 'searches=[{
    "collection": "docs",
[...]

PR #2817, opened 2026-03-05
fix: preserve vector search for zero-match phrase queries

alangmartini:
## Summary
- preserve vector search execution when phrase-query keyword matching returns zero candidates
- avoid early exit in phrase path when a vector query is present
- add regression coverage for phrase hybrid searches with no keyword matches (alpha 0.5 and alpha 0.0)

## Testing
- bazel build //:typesense-server
- bazel test --cache_test_results=no --test_output=all //:typesense-test --test_filter=CollectionVectorTest.TestHybridPhraseQueryFallbacksToVectorSearch --test_timeout=900
- bazel test --cache_test_results=no --test_output=all //:typesense-test --test_timeout=900 (suite has existing unrelated personalization test failures in this environment)

### DIFF
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -3619,7 +3619,7 @@ Option<bool> Index::search(std::vector<query_tokens_t>& field_query_tokens, cons
             return do_phrase_search_op;
         }
 
-        if (filter_result_iterator->approx_filter_ids_length == 0) {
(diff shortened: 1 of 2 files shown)


alangmartini:
Fixes #2816

kishorenc (approved):

Issue #2824, opened 2026-03-09
Stemming breaks curation filter application


YonoYing:
## Bug Description
When using the new `stem` option added to 30.1 for curations, setting `"stem": false` does not disable stemming for curations with a `filter_by` clause.

## Reproduction Steps
Minimal reproducible example: https://gist.github.com/YonoYing/8162ac55a007e8948a7030e77c5f4948

```bash
curation_set = {
  "rule": {
    "match": "contains",
    "query": "diabetes",
    "stem": false
  },
  "filter_by": "tag_ids:=[29266, 29267]",
  "remove_matched_tokens": true,
  "filter_curated_hits": false,
  "excludes": [],
  "includes": [],
  "stop_processing": true
}
```

## Expected vs Actual
**Expected behavior**  
The above curation should only return the two tag_ids in the `filter_by` clause

**Actual behavior**  
The above curation returns all stemmed matches for `diabetes`, including `diabet`

## Environment
- **Typesense version:** e.g. `v30.2.rc6`  
- **Operating system:** `Ubuntu 22.04 x86_64`, 

## Schema / Configuration

Include any relevant schema definitions, JSON payloads, or config files:

```json
{
  "name": "recipes",
  "fields": [
    {"name": "name", "type": "string", ... "stem": true},
    ...
  ],
  "curation_sets": ["recipe_tag_filters"]
}
```

PR #2826, opened 2026-03-10
fix stemming curations when search field is having them enabled

krunal1313:
## Change Summary

- Fix stemming curation when curation rule is having `stem:false` but search field is having `stem:true`
## PR Checklist


krunal1313:
Fixes #2824

Issue #2610, opened 2025-10-12
[Feature Request] Auto-bucket by text_match score similarity threshold


Clasyc:
**Describe your feature request**

Add a new bucketing option that groups search results based on the similarity of their `_text_match` scores, rather than fixed bucket sizes or counts. This would allow results with nearly identical text relevance to be sorted by popularity, while maintaining clear separation between results with meaningfully different relevance scores.

Proposed syntax example:
```json
{
  "sort_by": "_text_match(bucket_auto_size: 0.5):desc,popularity_score:desc"
}
```

Results would be grouped into buckets where all items within a bucket have `_text_match` scores within the specified threshold of each other. Within each bucket, results are then sorted by the secondary field (popularity_score).

**Please describe your use-case for the feature request**

We are building an e-commerce search experience where I need to balance text relevance with product popularity. The current `bucket_size` parameter creates problems when products have very similar text match scores.

Example: Searching for "wireless headphones" with `bucket_size: 5`:

| Result | Product | Text Match | Popularity |
|--------|---------|------------|------------|
| 1-5 | Various products | 24.8 | 25-120 |
| **6** | **Apple AirPods Max** | **24.7** | **950** |
| 7-10 | Various products | 24.7 | 40-200 |

The Apple AirPods Max (popularity: 950) is stuck at position 6 in Bucket 2, while products with popularity scores of 25-30 appear in positions 4-5 simply because they're in Bucket 1. The text match difference is only 0.1 points (24.8 vs 24.7), which is negligible.

With score-threshold bucketing (`bucket_threshold: 0.5`), all products with scores 24.7-24.8 would fall into the same bucket and be sorted by popularity, allowing the AirPods Max to rank higher where it belongs.

This is critical for e-commerce where popular products should surface when their text relevance is nearly identical to other results.

**Describe alternatives you've considered**

1. **Using `buckets` parameter** - Doesn't solve the problem; just divides results evenly regardless of score distribution
2. **Using smaller `bucket_size`** - Creates too much granularity and popularity has minimal effect
3. **Using larger `bucket_size`** - Mixes truly irrelevant results with relevant ones
4. **Manual re-ranking on application side** - Defeats the purpose of Typesense's built-in ranking capabilities

None of these alternatives allow grouping by actual score similarity.

**Additional context**

I'm not entirely sure if this approach is computationally feasible without sacrificing search latency. This might add overhead compared to the current fixed-size bucketing which can pre-determine bucket boundaries.

[...]

PR #2832, opened 2026-03-13
feat: add bucket_auto_size parameter for _text_match sorting

Rahik-Sikder:
## Add `bucket_auto_size` parameter for `_text_match` sorting
Adds a new `bucket_auto_size` parameter to `_text_match` sort expressions that groups results into variable sized buckets based on a score similarity threshold rather than fixed positional boundaries. The threshold is passed in as a percentage and is calculated using the `text_match` score of the first item in the bucket. 

Fixes issue #2610 , which claimed that two documents with nearly identical text match scores could end up in different buckets simply because they straddle a positional boundary, while two documents with wildly different relevance could land in the same bucket if the batch happens to be uniform. The issue originally proposed passing in a direct `text_match` value threshold, but I've implemented it as a percentage instead. Let me know if this solution isn't ideal, I've also added a comment on the original issue about it.

**Syntax Example:**
```
sort_by=_text_match(bucket_auto_size:0.25):desc,popularity:desc
```

Here, a threshold of `0.25` means any result within 25% of the top score in the current bucket is grouped with it. When a result falls outside that range, a new bucket starts. Secondary sort fields apply within each bucket.

**Compared to existing parameters:**
  - `buckets:N` - splits top N results into N equal-width positional buckets
  - `bucket_size:N` - each bucket holds exactly N consecutive results                                
  - `bucket_auto_size:T` -  bucket boundaries are determined by actual score differences, not position

  **Valid range:** `(0, 1]`. Zero and values above 1.0 are rejected with HTTP 400.

**Files changed:**
  - `include/field.h` -  added `float text_match_bucket_auto_size` to `sort_by` struct
  - `src/collection.cpp` - parsing in `validate_and_standardize_sort_fields`, bucketing logic in `search`                                                                                           
  - `test/collection_sorting_test.cpp` - correctness and validation tests

## PR Checklist

### DIFF
--- a/include/field.h
+++ b/include/field.h
@@ -622,6 +622,7 @@ struct sort_by {
     // for text_match score bucketing
     uint32_t text_match_buckets;
     uint32_t text_match_bucket_size;
+    float text_match_bucket_auto_size;
 
     uint32_t vector_search_buckets = 0;
     uint32_t vector_search_bucket_size = 0;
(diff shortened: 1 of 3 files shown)


Rahik-Sikder:
Just wondering if anyone has been able to look over this yet

Issue #2834, opened 2026-03-16
Streamed conversation responses silently drop tokens when SSE events span curl buffer boundaries


alangmartini:
## Bug Description

When using `conversation_stream=true` with `multi_search`, tokens are silently dropped from the streamed response. The reassembled text has missing words — e.g. `"You define Terraform using."` instead of `"You can define a Terraform monitor using Terraform."`.

The bug is in how `async_res_write_callback` parses SSE events from the upstream LLM. Each curl callback invocation applies this regex independently:

```cpp
std::regex data_regex("data: (.*?)\\n\\n");
```

This only matches SSE events that are **fully contained** within a single curl buffer delivery. Curl delivers data at arbitrary TCP segment boundaries. When an SSE event spans two curl callbacks, the partial tail in buffer N has no closing `\n\n` (not matched), and the partial head in buffer N+1 has no `data: ` prefix (not matched). Both halves are silently discarded.

There is no remainder buffer carried between callback invocations — each call to `async_res_write_callback` is stateless.

The same pattern exists in all provider implementations:

- `OpenAIConversationModel::async_res_write_callback` — `src/conversation_model.cpp:539`
- `CFConversationModel::async_res_write_callback`
- `vLLMConversationModel::async_res_write_callback`
- `AzureConversationModel::async_res_write_callback`

Additionally, when `nlohmann::json::parse()` throws on a partial JSON fragment, the catch block logs the error but leaves the `response` string unmodified — the raw upstream chunk gets forwarded to the client as-is.

## Reproduction Steps

A self-contained reproducer using a mock OpenAI server that deliberately splits SSE events across TCP writes.

The mock server (`mock_openai.py`) sends 9 tokens as SSE events but splits them across TCP writes so some `data: {...}\n\n` events span two curl buffer deliveries.

Execute the reproduce.sh attached with mock_openai.py in the same folder.

[mock_openai.py](https://github.com/user-attachments/files/26038302/mock_openai.py)
[reproduce.sh](https://github.com/user-attachments/files/26038303/reproduce.sh)

## Expected vs Actual

**Expected behavior**

All 9 tokens from the upstream LLM are delivered to the client via SSE:

```
data: {"conversation_id":"...","message":"You"}
data: {"conversation_id":"...","message":" can"}
data: {"conversation_id":"...","message":" define"}
data: {"conversation_id":"...","message":" a"}
data: {"conversation_id":"...","message":" server"}
data: {"conversation_id":"...","message":" monitor"}
[...]

PR #2836, opened 2026-03-17
fix: handle Gemini streamed responses across curl buffer boundaries

ozanarmagan:
## Change Summary
  Fixes Gemini conversation streaming when a streamed JSON object is split across curl callback boundaries.

  This change adds per-request buffering for Gemini stream chunks, incrementally extracts complete JSON objects from the streamed array response, and only
  parses full objects once they are complete. That prevents dropped tokens when an object spans multiple callback invocations.

  It also adds regression coverage for:
  - a Gemini object split across callbacks
  - array delimiter chunks arriving separately, including , and trailing ]

## PR Checklist

### DIFF
--- a/include/conversation_model.h
+++ b/include/conversation_model.h
@@ -9,6 +9,7 @@
 
 struct async_conversation_t {
     std::string response;
+    std::string stream_remainder;
     std::mutex mutex;
     std::condition_variable cv;
     long status_code;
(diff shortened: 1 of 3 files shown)


ozanarmagan:
Fixes #2834

kishorenc (approved):

PR #2837, opened 2026-03-17
fix: prevent token loss in streamed conversation responses across curl buffer boundaries

ozanarmagan:
## Change Summary
Fixes streamed conversation responses dropping tokens when upstream SSE events span curl callback boundaries.

This change adds per-request buffering for incomplete SSE fragments and updates the OpenAI, Cloudflare, vLLM, and Azure streaming parsers to process only complete SSE events. That prevents partial events from being silently discarded when curl delivers stream data across arbitrary buffer boundaries.

It also adds regression coverage for the split SSE events

Additionally, it tightens async completion signaling in the touched code paths by synchronizing ready updates before notifying the waiting thread.

## PR Checklist

### DIFF
--- a/include/conversation_model.h
+++ b/include/conversation_model.h
@@ -9,6 +9,7 @@
 
 struct async_conversation_t {
     std::string response;
+    std::string sse_remainder;
     std::mutex mutex;
     std::condition_variable cv;
     long status_code;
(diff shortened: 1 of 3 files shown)


ozanarmagan:
Fixes #2834

kishorenc (approved):

kishorenc (approved):

Issue #2760, opened 2026-02-04
Faceting on joined reference fields and facet_return_parent


forsdahl:
## Bug Description
When faceting on a joined reference field, `facet_return_parent` isn't working at all, even though this would be expected to work the same way as when using nested fields.

## Reproduction Steps
```bash
set -x

export TYPESENSE_API_KEY=xyz
export TYPESENSE_HOST=http://localhost:8108

docker stop typesense-repro 2>/dev/null
docker rm typesense-repro 2>/dev/null
rm -rf "$(pwd)"/typesense-data-dir-repro
mkdir "$(pwd)"/typesense-data-dir-repro

# Wait for Typesense to be ready
docker run -d -p 8108:8108 --name typesense-repro \
            -v"$(pwd)"/typesense-data-dir-repro:/data \
            typesense/typesense:30.1 \
            --data-dir /data \
            --api-key=$TYPESENSE_API_KEY \
            --enable-cors

# Wait till typesense is ready.
until curl -s -o /dev/null -w "%{http_code}" "$TYPESENSE_HOST/health" -H "X-TYPESENSE-API-KEY: ${TYPESENSE_API_KEY}" | grep -q "200"; do
  sleep 2
done

curl -s "$TYPESENSE_HOST/debug" \
       -H "X-TYPESENSE-API-KEY: ${TYPESENSE_API_KEY}" | jq


curl -s "$TYPESENSE_HOST/collections" \
       -X POST \
       -H "Content-Type: application/json" \
       -H "X-TYPESENSE-API-KEY: ${TYPESENSE_API_KEY}" \
       -d '
          {
             "name": "AttributeTypes",
             "fields": [
               {"name": "name", "type": "string" },
               {"name": "label", "type": "string" },
               {"name": "sort", "type": "int32" }
             ],
             "default_sorting_field": "sort"
           }' | jq

curl -s "$TYPESENSE_HOST/collections" \
       -X POST \
       -H "Content-Type: application/json" \
       -H "X-TYPESENSE-API-KEY: ${TYPESENSE_API_KEY}" \
       -d '
          {
             "name": "AttributeValues",
             "fields": [
               {"name": "value", "type": "string", "facet": true },
               {"name": "type_id", "type": "string", "reference": "AttributeTypes.id" },
               {"name": "sort", "type": "int32" }
             ],
             "default_sorting_field": "sort"
           }' | jq

[...]

PR #2849, opened 2026-03-25
support facet_return_parent for reference facets

krunal1313:
## Change Summary

This change fixes `facet_return_parent` for joined reference facets and aligns union-search validation with the runtime behavior already used for faceting.

### What changed

For facet serialization:
- parent return is no longer limited to nested local facets only
- parent return now also works for joined reference facets

For joined reference facets:
- the representative document is fetched from the referenced collection
- the sequence key is resolved from the referenced collection
- parent extraction is executed against the referenced collection

### Parent Matching Fix

The parent lookup helper now normalizes JSON values using facet-compatible string conversion for:
- strings
- signed integers
- unsigned integers
- floats
- booleans
- arrays of those values

This matters because joined facet values such as `product_price` are serialized as strings like `143` or `73.5`, and the parent finder must compare against the same representation.

Without this normalization, joined numeric facets can fail to locate the correct parent object even after the referenced collection lookup is fixed.

### Union Validation Fix
Union validation now:
- parses each `facet_by` entry using the actual collection for that search
- uses the parsed facet field name for `facet_return_parent` consistency checks
- compares parsed facet shape instead of raw `facet_by` text when checking facet uniformity

The parsed facet signature includes:
- leaf field name
- referenced collection name
- range configuration
- alpha sorting flag
- sort order
- sort field
- top-k flag

This keeps validation aligned with the runtime logic while remaining narrow in scope.

### Alias vs Canonical Joined Facets

Union validation now treats these as the same underlying joined facet shape when they resolve to the same referenced collection:
- `$Countries(name)`
- `$Countries_alias(name)`

This is handled by comparing parsed facet signature rather than raw `facet_by` text.

## Known Limitation

A known limitation remains:
- distinct facets that share the same leaf field name are still ambiguous under current runtime semantics

Examples:
- local `name` and `$Countries(name)`
- `$Countries(name)` and `$Brands(name)`

This PR does not attempt to solve that because doing so would require a broader selector/identity design change. The current implementation keeps runtime and validation aligned without expanding the API surface.

## PR Checklist

### DIFF
--- a/include/collection_manager.h
+++ b/include/collection_manager.h
@@ -60,7 +60,8 @@ private:
         return "";
     }
 
-    static Option<bool> validate_facet_params(const std::vector<collection_search_args_t>& coll_searches);
+    static Option<bool> validate_facet_params(const std::vector<collection_search_args_t>& coll_searches,
+                                              const std::vector<std::shared_ptr<Collection>>& collections);
 
 public:
     static constexpr const size_t DEFAULT_NUM_MEMORY_SHARDS = 4;
--- a/src/collection.cpp
+++ b/src/collection.cpp
@@ -4888,17 +4888,31 @@ nlohmann::json Collection::get_parent_object(const nlohmann::json& parent, const
                                  const std::vector<std::string>& field_path, size_t field_index,
                                  const std::string& val) {
     if(field_index == field_path.size()) {
-        std::string str_val;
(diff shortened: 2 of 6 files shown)


krunal1313:
Fixes #2760

kishorenc (approved):

PR #2852, opened 2026-03-26
fix(joins): allow null optional nested references in object arrays on update

tharropoulos:
## Change Summary
  - allow optional nested reference fields in object arrays to skip null values during  update and async reference handling
  - accept signed and unsigned json integers within the uint32 range for nested reference
    helper arrays
  - add tests for nested reference updates, code-based references and mixed null/unresolved async cases

## PR Checklist

### DIFF
--- a/src/validator.cpp
+++ b/src/validator.cpp
@@ -1,6 +1,17 @@
 #include "validator.h"
 #include "field.h"
 
+namespace {
+bool is_valid_uint32_json_integer(const nlohmann::json& value) {
+    if (!value.is_number_integer() && !value.is_number_unsigned()) {
+        return false;
+    }
+
+    const auto signed_value = value.get<int64_t>();
+    return signed_value >= 0 && static_cast<uint64_t>(signed_value) <= UINT32_MAX;
+}
+}
+
 Option<uint32_t> validator_t::coerce_element(const field& a_field, nlohmann::json& document,
                                        nlohmann::json& doc_ele,
                                        const std::string& fallback_field_type,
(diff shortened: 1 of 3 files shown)


kishorenc:
@happy-san Please review.

src/validator.cpp
```diff
@@ -1,6 +1,17 @@
 #include "validator.h"
 #include "field.h"
 
+namespace {
```
happy-san, line 4: @tharropoulos Is there a reason for having `namespace` here?

test/collection_join_test.cpp
```diff
@@ -1117,6 +1117,245 @@ TEST_F(CollectionJoinTest, IndexDocumentHavingReferenceField) {
...
+    auto get_op = items_collection->get("item-001");
+    ASSERT_TRUE(get_op.ok());
+    auto doc = get_op.get();
+    ASSERT_EQ("Updated Test Item - Reference Still Null", doc["title"]);
+    ASSERT_EQ(1, doc.count("relatedassetsdata"));
+    ASSERT_EQ(1, doc["relatedassetsdata"].size());
+    ASSERT_EQ(0, doc["relatedassetsdata"][0].count("childobjectid"));
+}
+
+TEST_F(CollectionJoinTest, UpdateOptionalNestedReferenceInObjectArrayViaPatchPathOnCodeField) {
```
happy-san, line 1193: This test passes even without the changes in this PR. Could you check if we're testing the change correctly?

src/validator.cpp
tharropoulos: Fair point. The unnamed namespace is only there to keep `is_valid_uint32_json_integer` file specific. Should I just change it to a static for consistency?

happy-san: I'll let @kishorenc take that call.

test/collection_join_test.cpp
tharropoulos: I tried tightening it a few ways to mirror the failing cases more closely, like switching to the single-document UPDATE path, changing null -> valid code reference, and adding multiple object-array slots so the helper field had to record a non-zero object index, but it still passed without the fix. 

At this point it seems this code-field variant is not targeting the broken path the PR fixes, while the two tests that do fail without the seem to be the real regressions.

 So I agree the current test is not validating the change correctly, and I'd treat it as general coverage rather than evidence for the fix unless you had a specific pre-fix repro in mind for that variant.


EDIT: readability changes

happy-san: I think I found the reason, could you check if this test hits the `else if (optional && object_array[i].at(keys[1]).is_null())` condition at line 278.

tharropoulos: It is

happy-san: Okay. Let's keep this test as is. The issue was only with the id field reference in an object array then.

happy-san:
@kishorenc apart from the `namespace`, the rest seems good to me.

tharropoulos:
@kishorenc any updates here?a

Issue #2846, opened 2026-03-24
Vector_search query produces incorrect facet data with group_by


monotype-vikas:
## Bug Description
On a vector search against a collection with joins to mtc_styles_data, facet_counts are inconsistent depending on whether group_by is used:

Without group_by: for a restrictive filter_by that includes joined fields (e.g. fdy_name, cvp_width_value, cvp_height_value inside $mtc_styles_data(...)), facet_counts are empty (counts: [], stats.total_values: 0 for all facet_by fields), consistent with no matches in the result set.

With the same filter_by, vector_query, facet_by, and per_page — only adding group_by (e.g. on family_id): facet_counts become non-empty, including a bucket for fdy_name = Inviline (and other facets with non-zero counts).
group_by should only change how hits are grouped/paginated, not whether the underlying filtered + vector-qualified document set is empty. Facet results should be the same for identical filter_by + vector_query regardless of group_by.

A concrete label we used while debugging: Inviline appears in fdy_name facets with group_by when the without-group_by run returns empty facets for the same query.

## Reproduction Steps

**search with group_by**
```json
{
  "collection": "mtc_font_embeddings_data",
  "q": "*",
  "filter_by": "$mtc_styles_data(fdy_name:= 'Inviline' && cvp_width_value:=300 && cvp_height_value:=100)",
  "vector_query": "body_embeddings:([0.752036,-0.797508,0,-0.86267,0.76113,0.684734,-0.718115,-0.397804,-0.716131,0.923868,0.80179,0.881492,-0.870096,-0.738616,0.788121,0.866227], k: 2000, distance_threshold: 1.5)",
  "per_page": 250,
  "page": 1,
  "include_fields": "style_id,family_id,$mtc_families_data(id,family_url_key,image_urls,name_en,title_en),$mtc_styles_data(id,font_name,font_url_key,font_nids,font_ps_names,md5,font_pim_style_id,is_custom,is_var,is_default,is_imported_font,derelease_date,render_md5,font_mat_num,tags_font_en,$mtc_foundries_data(handle,logo,name_en,title_en),$mtc_list_style_mappings_data(list_id,metadata,style_id))",
  "exclude_fields": "heading_embeddings,body_embeddings",
  "facet_by": "$mtc_styles_data(fdy_name),$mtc_styles_data(font_advanced_typography),$mtc_styles_data(font_langs_en),$mtc_styles_data(tags_font_en)",
  "max_facet_values": 50,
  "include_facets": true,
[...]

PR #2865, opened 2026-04-02
Fix grouped vector search stale facets and counts

krunal1313:
## Change Summary

Grouped search runs in two passes.

The original implementation reused first-pass grouped state when building grouped results. That was fine for classic grouped search, but it broke for vector and hybrid queries because the first pass was not final-qualified by vector distance or hybrid reranking.

There were two separate issues:

1. First-pass grouped state could leak into the final response.
2. First-pass group discovery skipped vector logic entirely, so vector-qualified groups could be missing from the candidate group set.

### Fix
The fix keeps grouped search as a two-pass flow, but makes the responsibilities clearer:

1. The first grouped pass uses isolated temporary state.
2. The first grouped pass is vector-aware, so candidate groups reflect vector-qualified matches.
3. The final grouped response is derived from the final pass only for grouped vector queries.
4. Non-vector grouped queries keep their historical approximate count behavior.

### Cleanup

The first-pass temporary state has been moved into a private helper on `Index`:

- [include/index.h](/home/krunal/typesense/include/index.h#L404)

This keeps the `run_search` grouped flow smaller and avoids a long block of inline `vector`, raw pointer, and `unique_ptr` declarations in the middle of the function.

The helper's `take_ownership()` method adopts the raw `Topster*` pointers returned by `search(...)` into `unique_ptr` so the temporary grouped pass state is cleaned up automatically.
## PR Checklist

### DIFF
--- a/include/index.h
+++ b/include/index.h
@@ -925,6 +925,19 @@ public:
                                  const bool& is_group_by_first_pass,
                                  std::set<uint32_t>& group_by_missing_value_ids) const;
 
+    void process_grouped_vector_results_hnsw(filter_result_iterator_t* filter_result_iterator_no_groups,
+                                             const vector_query_t& vector_query,
+                                             hnsw_index_t* field_vector_index,
+                                             VectorFilterFunctor& filter_functor,
+                                             size_t initial_k,
+                                             size_t fetch_size,
+                                             size_t group_limit,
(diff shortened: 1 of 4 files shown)


krunal1313:
Fixes #2846

kishorenc (approved):

Issue #2858, opened 2026-03-29
Bug: Conversational Search with Cloudflare Workers AI fails — `type must be string, but is null`


PavieOlivier:
## Description

Conversational Search (RAG) with Cloudflare Workers AI as the LLM provider fails with `"Got malformed response from Cloudflare API."`, even though the model **correctly generates and streams the full answer**.

The root cause is in `conversation_model.cpp:978` — Typesense expects the `response` field in every SSE chunk to be a string, but Cloudflare Workers AI sends `"response": null` in the final usage/summary chunk of its streaming response.

## Typesense Version

- **v30.1** (self-hosted via Docker: `typesense/typesense:30.1`)

### 1. Create a conversation model using Cloudflare Workers AI

```bash
curl "http://localhost:8108/conversations/models" \
  -X POST \
  -H "Content-Type: application/json" \
  -H "X-TYPESENSE-API-KEY: ${TYPESENSE_API_KEY}" \
  -d '{
    "id": "my-chat-model",
    "model_name": "cloudflare/@cf/meta/llama-3.1-8b-instruct",
    "api_key": "<CLOUDFLARE_API_TOKEN>",
    "account_id": "<CLOUDFLARE_ACCOUNT_ID>",
    "history_collection": "conversation_store",
    "max_bytes": 16384,
    "ttl": 3600,
    "system_prompt": "You are a helpful assistant. Be concise."
  }'
```

### 2. Run a conversational search query

```bash
curl "http://localhost:8108/multi_search?conversation=true&conversation_model_id=my-chat-model&q=pizza" \
  -X POST \
  -H "Content-Type: application/json" \
  -H "X-TYPESENSE-API-KEY: ${TYPESENSE_API_KEY}" \
  -d '{
    "searches": [{
      "collection": "my_collection",
      "query_by": "embedding",
      "exclude_fields": "embedding",
      "per_page": 3
    }]
  }'
```

### 3. Observe the error

**API response:**
```json
{"message": "Got malformed response from Cloudflare API."}
```

**Server log error:**
```
E20260329 06:34:29.511289  conversation_model.cpp:978] [json.exception.type_error.302] type must be string, but is null
```

## Root Cause Analysis

The Cloudflare Workers AI streaming response (SSE format) looks like this:

```
data: {"response":"Hello","p":"..."}

data: {"response":" world","p":"..."}

data: {"response":null,"usage":{"prompt_tokens":159,"completion_tokens":98,"total_tokens":257}}

[...]

PR #2871, opened 2026-04-12
fix: ignore null response chunks in stream response parsing

ozanarmagan:
## Change Summary
For CF conversation models with streaming, any response chunk with `null` or undefined response value causes whole stream to fail. Instead, this PR proposes ignoring the such chunks.

## PR Checklist


ozanarmagan:
#2858

kishorenc (approved):

Issue #2862, opened 2026-04-01
Facet counts are incorrect when group_by is combined with join


dejonghee:
## Bug Description
When combining a 'group_by' with a 'join', the facet count of the joined fields are incorrect.

## Reproduction Steps
The following repro can be used to simulate the issue:

```
@hostname = localhost:46675
@apiKey = typesense-api-key

### Start with a clean slate by deleting existing collections.

DELETE http://{{hostname}}/collections/price
X-TYPESENSE-API-KEY: {{ApiKey}}

###

DELETE http://{{hostname}}/collections/variants
X-TYPESENSE-API-KEY: {{ApiKey}}

###

DELETE http://{{hostname}}/collections/stock
X-TYPESENSE-API-KEY: {{ApiKey}}

### Create Variants Collection.

POST http://{{hostname}}/collections
X-TYPESENSE-API-KEY: {{apiKey}}
Content-Type: application/json

{
  "name": "variants",
  "fields": [
    {"name": "product_id", "type": "string", "reference": "productsalias.id", "async_reference": true, "facet": true},
    {"name": "color", "type": "string", "facet": true, "sort": true}
  ],
  "metadata": {"revision": "8a8793bf-7592-48fc-ac14-ab0896bc9ba0"}
}

### Create Alias for Variants.

PUT http://{{hostname}}/aliases/variantsalias
X-TYPESENSE-API-KEY: {{apiKey}}
Content-Type: application/json

{
  "collection_name": "variants"
}

### Create Price Collection.

POST http://{{hostname}}/collections
X-TYPESENSE-API-KEY: {{apiKey}}
Content-Type: application/json

{
  "name": "price",
  "fields": [
    {"name": "variant_id", "type": "string", "reference": "variantsalias.id"},
    {"name": "listPrice", "type": "float", "facet": true, "sort": true},
    {"name": "salePrice", "type": "float", "facet": true, "sort": true},
    {"name": "currency", "type": "string" },
    {"name": "priceListId", "type": "string" }
  ]
}

### Create Alias for Price.

PUT http://{{hostname}}/aliases/pricealias
X-TYPESENSE-API-KEY: {{apiKey}}
Content-Type: application/json

{
  "collection_name": "price"
}

### Create Stock Collection.

POST http://{{hostname}}/collections
X-TYPESENSE-API-KEY: {{apiKey}}
Content-Type: application/json

{
  "name": "stock",
  "fields": [
[...]

PR #2877, opened 2026-04-14
Fix grouped join facet counts

krunal1313:
### Issue
Joined facet counts were being computed after parent group information was already lost. Because of that, grouped joined facets could collapse to count `1`.

### Fix
- Added a reference facet context that keeps:
  - referenced doc ids
  - parent group ids for each referenced doc
- Updated grouped joined faceting to count distinct parent groups instead of regrouping inside the referenced collection
- Fixed joined `facet_query` so it works against referenced facet fields
- Unified reference facet collection into one path
- Added regression tests for:
  - grouped joined string facets
  - grouped joined bool facets
  - grouped joined range facets
  - grouped joined `facet_query`
## PR Checklist


krunal1313:
Fixes #2862

PR #2881, opened 2026-04-21
fix: correct exhaustive faceting across multiple nested join paths

tharropoulos:
## Change Summary
- add a regression test for facet_by with multiple nested join paths in exhaustive mode to verify include_fields still resolves both join paths correctly in the same setup
- fix reference facet batching in index::search so batches are built per thread rather than per referenced collection

## PR Checklist


happy-san (approved):
Looks good!

Pernett98:
I love that you fix this 👍🏼

Pernett98:
@tharropoulos @happy-san is there anything that I can do to help on this PR ? I would love to have this fix

Issue #2890, opened 2026-04-24
Natural Language Models not loading on snapshot recover.


alangmartini:
## Bug Description

Natural Language Search models vanish from the API after a Raft snapshot is reloaded into a node. The `$NLSP_*` rows survive on disk in RocksDB, but the in-memory `models` map in `NaturalLanguageSearchModelManager` stays empty, so `GET /nl_search_models` returns `[]` and any `nl_query=true` search fails with "model not found".

Root cause: `NaturalLanguageSearchModelManager::init` is called exactly once, **before** Raft replication starts, in `src/typesense_server_utils.cpp:659`. When a node boots with an existing snapshot in its state dir or receives one from a peer, `ReplicationState::on_snapshot_load` (`src/raft_server.cpp:705`) runs `store->reload(...)` followed by `init_db()`. `init_db()` (`src/raft_server.cpp:663-703`) re-initializes `CollectionManager` (line 666), `ConversationModelManager` (line 678), and `PersonalizationModelManager` (line 695), but has no equivalent call for `NaturalLanguageSearchModelManager`. The NL manager's in-memory cache therefore stays empty for the life of the process.

The bug is **100% deterministic**. Verified across 5 back-to-back restarts with the reproducer below, polling `/nl_search_models` 7 times across a 30-second post-reload observation window each run. The API returned `[]` on every single poll (0/35 attempts recovered).

## Reproduction Steps

Because the bug lives in `on_snapshot_load`, a pure `curl`-only repro isn't possible — the script below uses `docker` + `curl` to stand up a single-node HA Typesense plus a tiny OpenAI-compatible mock so `validate_vllm_model` accepts the two test models without any real LLM credentials. Every Typesense API interaction is a plain `curl`. Save both files side by side, `chmod +x reproduce.sh`, run `./reproduce.sh`.

### `mock-llm-server.py`

```python
from http.server import BaseHTTPRequestHandler, HTTPServer
import json


class Handler(BaseHTTPRequestHandler):
    def log_message(self, fmt, *args):
        pass

    def do_POST(self):
        length = int(self.headers.get("Content-Length", 0))
        if length:
            self.rfile.read(length)
        body = {
            "id": "mock",
            "object": "chat.completion",
            "created": 0,
            "model": "mock",
            "choices": [
                {
                    "index": 0,
                    "message": {"role": "assistant", "content": "{\"q\":\"*\"}"},
                    "finish_reason": "stop",
                }
            ],
[...]

PR #2891, opened 2026-04-25
fix reloading nl models from snapshot recover

krunal1313:
#### Issue
Natural language search models were being initialized only once during server startup, before Raft snapshot restore completed.

That meant:
- NL model metadata could still exist on disk in RocksDB
- collections, conversation models, and personalization models would be reloaded after snapshot restore
- but `NaturalLanguageSearchModelManager` would keep an empty in-memory cache for the lifetime of the process

#### Changes
- Moved NL model initialization into the Raft reload path
- Removed the old one-time startup init
- Added test

## PR Checklist

### DIFF
--- a/src/collection_manager.cpp
+++ b/src/collection_manager.cpp
@@ -16,6 +16,7 @@
 #include "core_api_utils.h"
 #include "synonym_index_manager.h"
 #include "curation_index_manager.h"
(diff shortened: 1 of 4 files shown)


krunal1313:
Fixes #2890

kishorenc (approved):

PR #2901, opened 2026-05-03
Updated so RAG/conversations with Azure APIs don't timeout

J-Cat:
Updated so RAG/conversations with Azure APIs don't timeout after 4 seconds but instead timeout after 60 seconds like the other providers.

## Change Summary
Updated AzureConversationModel::get_answer to use HttpProxy::default_timeout_ms instead of leaving blank which defaults it to 4 seconds (vs 60 seconds).  This makes the Azure model/get answer in line with the other providers that use the HttpProxy::default_timeout_ms value that is 60s.

## PR Checklist

- [ x] I have read and signed the [Contributor License Agreement](https://forms.gle/PZyiY5N2GDQU8GsV9).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ typesense-data
 /dist
 /tmp
 *.tar.gz
+.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -20,4 +20,3 @@ typesense-data
 /dist
 /tmp
 *.tar.gz
-.vscode
--- a/src/conversation_model.cpp
+++ b/src/conversation_model.cpp
@@ -2067,7 +2067,7 @@ Option<std::string> AzureConversationModel::get_answer(const std::string& contex
     headers["api-key"] = api_key;
     headers["Content-Type"] = "application/json";
 
-    long status_code = HttpClient::post_response(url, request_body.dump(), response, res_headers, headers);
+    long status_code = HttpClient::post_response(url, request_body.dump(), response, res_headers, headers, HttpProxy::default_timeout_ms);
 
     if (status_code != 200) {
         return Option<std::string>(status_code, "Failed to get response from Azure API: " + response);


.gitignore
```diff
@@ -20,3 +20,4 @@ typesense-data
 /dist
 /tmp
 *.tar.gz
+.vscode
```
tharropoulos: This can be left out

src/conversation_model.cpp
```diff
@@ -2067,7 +2067,7 @@ Option<std::string> AzureConversationModel::get_answer(const std::string& contex
     headers["api-key"] = api_key;
     headers["Content-Type"] = "application/json";
 
-    long status_code = HttpClient::post_response(url, request_body.dump(), response, res_headers, headers);
+    long status_code = HttpClient::post_response(url, request_body.dump(), response, res_headers, headers, HttpProxy::default_timeout_ms);
```
tharropoulos, line 2070: Would it be valuable to add this under every model provider instead of just Azure?

.gitignore
J-Cat: Yes, just added it in since I was working in VSCode and didn't want my local VSCode config to be part of the pull request.

src/conversation_model.cpp
J-Cat: Every other model does use HttpProxy::default_timeout_ms in their call.  Was just the Azure provider that used HttpClient instead of RemoteEmbedder that had this issue.  (I checked the other ones as I thought the same thing)

tharropoulos: Oh, sorry.

.gitignore
tharropoulos: Could you force push it out? It's outside of the scope of this PR

J-Cat:
Yes, just to understand correctly, you want me to remove the change to .gitignore and re-issue the pull request?  (not sure exactly what you mean by “force push it out”)

J-Cat:
Removed the changes to .gitignore

.gitignore
J-Cat: Removed it.

tharropoulos:
You can amend your commit and force push to remove the changes from .gitignore, just so it doesn't interfere with git history

J-Cat:
I’ve removed it (no longer shows up in the changed files)

tharropoulos:
Amending a commit and using `git push fork --force-with-lease` will remove those gitignore changes.  Up to @kishorenc

PR #2904, opened 2026-05-04
fix: Query string not re-tokenized per field when using field-level token_separators/symbols_to_index  #2828

### LINKED ISSUE #2828: Query string not re-tokenized per field when using field-level token_separators/symbols_to_index with multiple query_by fields
## Bug Description

When using field-level `token_separators` and/or `symbols_to_index` with multiple `query_by` fields, Typesense tokenizes the query string **once**, using the rules of the **first (or highest-weighted) field** in the `query_by` list. The query is not re-tokenized when searching subsequent fields that have different tokenization rules.

This causes a significant reduction in hit count on lower-ranked fields when the first field's tokenization rules are incompatible with how those fields were indexed.

The v30.1 release notes reference a fix for "query sub-tokenization to respect field-level `symbols_to_index` and `token_separators` configuration" — but the bug is still present in v30.1 when mixing fields with different tokenization configs in a single `query_by`.

## Steps to Reproduce

**1. Create a collection with two fields that have different tokenization rules:**
```bash
curl -X POST http://localhost:8108/collections \
  -H "X-TYPESENSE-API-KEY: xyz" \
  -d '{
    "name": "products",
    "fields": [
      {
        "name": "sku",
        "type": "string",
        "symbols_to_index": ["-"],
      },
      {
        "name": "title",
        "type": "string",
        "token_separators": ["-", ".", "/"]
      }
    ]
  }'
```

**2. Index two documents:**
```bash
curl -X POST "http://localhost:8108/collections/products/documents/import" \
  -H "X-TYPESENSE-API-KEY: xyz" \
  -H "Content-Type: text/plain" \
  -d '{"id":"1","sku":"ABC-1234","title":"Vintage Wooden Chair"}
{"id":"2","sku":"XYZ-5678","title":"Mid-Century Modern Steel Desk"}'
```

**3. Search with `sku` as the first `query_by` field:**
```bash
curl "http://localhost:8108/collections/products/documents/search" \
  -H "X-TYPESENSE-API-KEY: xyz" \
  -G \
  --data-urlencode "q=wooden" \
  --data-urlencode "query_by=sku,title"
```

## Expected Behavior

`"Vintage Wooden Chair"` (document 1) should be returned because `"wooden"` matches the `title` field. The query `"wooden"` contains no characters in `sku`'s `token_separators`, so its tokenization is neutral — but even with a query like `"wooden-desk"`, the `title` field should be searched using its own tokenization rules (no splitting on `-`), not `sku`'s.

## Actual Behavior

Searching `query_by=sku,title` returns **fewer results** than searching `query_by=title` alone for the same query. The query string is tokenized using `sku`'s `token_separators` and `symbols_to_index` rules, and those tokens are then used verbatim when searching `title` — which was indexed without those separators.

This is demonstrable by running these two searches and comparing `found`:
```bash
# Search 1: sku first
curl "...?q=wooden-desk&query_by=sku,title"

# Search 2: title only
curl "...?q=wooden-desk&query_by=title"
```

Search 1 returns fewer results than Search 2 because `"wooden-desk"` is split into `["wooden", "desk"]` by `sku`'s `-` separator, and those tokens are then used against `title` — which indexed `"wooden-desk"` as a single token.

## Root Cause (Hypothesis)

Typesense appears to perform a single query tokenization pass before fanning the search out across `query_by` fields. The tokenization uses the rules of the first field in the list. Correct behavior would be to re-tokenize the query independently for each field, using that field's own `token_separators` and `symbols_to_index` configuration.

## Impact

This makes field-level `token_separators` and `symbols_to_index` — introduced in v28 and present in v30.1 — **effectively unusable in practice** for any schema that mixes identifier fields (e.g., SKUs, part numbers, barcodes) with natural-language fields (e.g., title, description, tags) in a single `query_by`.

## Workaround Attempted

Using `multiSearch` with the `query_by` fields split across two separate searches avoids the tokenization bleed. However, this workaround is blocked by the fact that `union` mode does not support `facet_by` (see related issue #2829), making it impractical for real-world search UIs that require faceted navigation.

## Environment

- **Typesense version:** 30.1
- **Deployment:** Typesense Cloud
- **OS:** macOS
- **Client library & version:** `typesense-js v1.3.0`

mdhaduk:
closes [#2828](https://github.com/typesense/typesense/issues/2828)

## Problem

When `query_by` spans multiple fields with different `symbols_to_index` or `token_separators` configurations, Typesense tokenized the query once using the first field’s rules and reused those tokens for every subsequent field.

This produced the wrong token set for fields that index content differently.

## Root Cause

In `collection.cpp`, `parse_search_query` was called once with `field[0]`’s:

- locale
- stemmer
- `symbols_to_index`
- `token_separators`

The result was then copied into every `field_query_tokens[i]` slot.

All downstream search paths used this single token set for all fields.

### `src/collection.cpp`

Re-tokenize the query for each field individually.

Field `0` is unchanged. Fields `i > 0` each get their own `parse_search_query` call using their own:

- locale
- stemmer
- `symbols_to_index`
- `token_separators`

### `src/index.cpp`

Updated three search paths:

#### 1. Primary path

Fields are grouped by their resulting token set using `field_group_t`.

Fields that produce identical tokens share a group and are passed together to `fuzzy_search_fields`, preserving cross-field matching.

Fields with different token sets get separate calls with their own tokens.

This ensures `orig_num_tokens` is always coherent with the fields being searched, so scoring is not artificially penalized.

#### 2. Space-as-typos fallback

When the primary search returns zero results and `split_join_tokens=fallback`, the resolved token set is now searched per group rather than broadcast to all fields via a single call.

#### 3. Drop-tokens path

`all_queries` used to be built from `field_query_tokens[0]` only. Truncated subsets were then searched across all fields with `sku`-style tokens that other fields never indexed.

The fix builds `group_all_queries` per group, so each group drops tokens from its own `N`-token representation.

## Why not just union all tokens and pass all fields once?

`fuzzy_search_fields` treats its token list as the single correct query representation for all fields given to it.

`orig_num_tokens` is set to the token count of that list.

With a union like (Referring to example in original issue [here](https://github.com/typesense/typesense/issues/2828)):

```txt
["wooden-desk", "wooden", "desk"]
```

a document perfectly matching title scores 2/3, while a perfect sku match scores 1/3.

Both are penalized relative to what they should be.

Grouping ensures each call has a coherent token count for accurate scoring.

## Testing

Extended PerFieldQueryRetokenization to cover the regression, field-order invariance, cross-field correctness, and the three-field tokenization case.

### DIFF
--- a/src/collection.cpp
+++ b/src/collection.cpp
@@ -2805,7 +2805,29 @@ Option<bool> Collection::init_index_search_args(collection_search_args_t& coll_a
 
         for(size_t i = 1; i < weighted_search_fields.size(); i++) {
             field_query_tokens.emplace_back(query_tokens_t{});
-            field_query_tokens[i] = field_query_tokens[0];
+            auto search_field_i = search_schema.at(weighted_search_fields[i].name);
+            std::vector<std::string> field_q_include_tokens;
+            std::vector<std::string> field_q_unstemmed_tokens;
+
+            parse_search_query(query, field_q_include_tokens, field_q_unstemmed_tokens,
+                               field_query_tokens[i].q_exclude_tokens,
+                               field_query_tokens[i].q_phrases,
+                               search_field_i.locale, pre_segmented_query, stopwords_set,
+                               search_field_i.get_stemmer(),
+                               search_field_i.symbols_to_index,
+                               search_field_i.token_separators);
+
+            for(size_t j = 0; j < field_q_include_tokens.size(); j++) {
+                auto& token = field_q_include_tokens[j];
+                field_query_tokens[i].q_include_tokens.emplace_back(
+                    j, token, (j == field_q_include_tokens.size() - 1), token.size(), 0);
+            }
+
+            for(size_t j = 0; j < field_q_unstemmed_tokens.size(); j++) {
+                auto& token = field_q_unstemmed_tokens[j];
+                field_query_tokens[i].q_unstemmed_tokens.emplace_back(
+                    j, token, (j == field_q_include_tokens.size() - 1), token.size(), 0);
+            }
         }
     }
 
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -4038,24 +4038,76 @@ Option<bool> Index::search(std::vector<query_tokens_t>& field_query_tokens, cons
             }
         }
 
-        auto fuzzy_search_fields_op = fuzzy_search_fields(the_fields, field_query_tokens[0].q_include_tokens, {}, match_type,
-                                                          excluded_result_ids, excluded_result_ids_size,
-                                                          filter_result_iterator, curated_ids_sorted,
-                                                          sort_fields_std, num_typos,
-                                                          searched_query_tokens,
-                                                          qtoken_set, topster, groups_processed,
-                                                          all_result_ids, all_result_ids_len,
-                                                          group_limit, group_by_fields, group_missing_values, prioritize_exact_match,
-                                                          prioritize_token_position, prioritize_num_matching_fields,
-                                                          query_hashes, token_order, prefixes,
-                                                          typo_tokens_threshold, exhaustive_search,
-                                                          max_candidates, min_len_1typo, min_len_2typo,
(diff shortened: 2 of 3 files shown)


mdhaduk:
Any update on this?

PR #2906, opened 2026-05-04
Implement cgroups conditions memory handling

### LINKED ISSUE #1715: Typesense reports and uses host machine memory stats when running in Docker
## Description

Typesense uses available memory stats to report it on `typesense_memory_*` metrics and for throttling writes with `memory-used-max-percentage` settings. But in docker container, where resource limits [are set via cgroups](https://docs.docker.com/config/containers/resource_constraints/), typesense reads the available memory stats from the host machine. It leads to OOM kills even with set up `memory-used-max-percentage` and incorrect metrics reporting.

The same issue is for Kubernetes environment with `resources:limits:` where nodes are usually huge servers. 

Side note: it is also true for CPU count which is used to calculate default values for some settings, but IMO it is not so important.

## Steps to reproduce

https://gist.github.com/b0g3r/2fcf88772cddf4bd4630a48bc1d63073

stafot:
## Fix: cgroup-aware memory guard for Docker/Kubernetes environments

Closes #1715

### Problem

The existing memory guard (`cached_resource_stat_t`) determines whether to reject writes by comparing used memory against the node's total RAM (via `sysinfo()` / `/proc/meminfo`). In containerised environments (Docker, Kubernetes) this is the wrong signal: the node may have 64 GB of free RAM while a pod is moments away from hitting its memory limit and being OOM-killed. The guard would never fire, and the process would be killed by the container runtime instead.

### What this PR does

Adds cgroup-aware memory reporting so that `get_memory_total_bytes()` and `get_memory_used_bytes()` return **container-scoped** values when running inside a cgroup, rather than node-level values.

- **cgroups v2** — reads `memory.max` (limit) and `memory.current` (usage) from `/sys/fs/cgroup/`
- **cgroups v1** — reads `memory.limit_in_bytes` and `memory.usage_in_bytes` from `/sys/fs/cgroup/memory/`; the well-known unlimited sentinel (`0x7FFFFFFFFFFFF000`) is treated as "no limit" and falls back to node-level stats
- Falls back to the existing `sysinfo()` / `/proc/meminfo` path when no cgroup limit is detected (bare-metal, macOS, unlimited cgroup)

With this change, the memory guard correctly tracks the pod's own memory budget and can reject writes before the pod reaches its limit, preventing OOM kills.

### Known limitation

The value reported by `memory.current` (cgroups v2) / `memory.usage_in_bytes` (cgroups v1) includes kernel page cache and other reclaimable memory that Kubernetes does not count towards the pod's Working Set. In practice this means the guard's reported usage can be significantly higher than what `kubectl top pod` shows — in our testing roughly **2.6 GB reported vs ~1.5 GB shown by Kubernetes** on the same workload.

As a result, the guard may reject some writes earlier than strictly necessary, which is not ideal for cost efficiency. However, this is a deliberate trade-off: **premature write rejection is far preferable to an OOM kill**, which terminates the process immediately and can corrupt in-flight operations.

A follow-up can refine the usage metric (e.g. using `memory.stat` → `anon` + `file_mapped` to better approximate the Kubernetes Working Set), but that is out of scope here.

### Testing

- Added unit tests covering cgroup v1, cgroup v2, the v1 unlimited sentinel, and the no-cgroup fallback path (`test/system_metrics_test.cpp`)
- Test fixture files added under `test/resources/` for both cgroup versions

### Deployment notes

No configuration changes required. The cgroup paths are the standard Linux defaults and will be detected automatically in any container runtime that mounts the cgroup filesystem (Docker, containerd, etc.).

---

### A note on collaboration

We rely heavily on Typesense in containerised environments and this functionality is important for our production reliability. That said, our goal is not to maintain a long-lived fork — we want this to live in upstream Typesense so the whole community benefits.

If maintainers see room for improvement (different usage metric, API design, edge cases we missed), we are fully open to iterating on this. Happy to address review feedback, rework the approach entirely if there is a preferred direction, or split this into smaller PRs if that helps the review process. We just want to land something solid upstream rather than carry a private patch indefinitely.

## PR Checklist


stafot:
@kishorenc When you have some time could you please take a look and review this work. As noted feel free to suggest refactor as needed.

krunal1313:
Hi @stafot 
This looks good, but cgroup files are being read from fixed paths only. In a lot of setups the process is in a nested cgroup, so `memory.max` /` memory.current` live under the path from `/proc/self/cgroup`, not directly under `/sys/fs/cgroup`.

If that happens, this falls back to host memory and the container memory guard still won’t trigger. Might be worth resolving the current cgroup path and adding a test for that case.

stafot:
Hi @krunal1313, thanks for the review and the sharp catch! Actually, after committing my initial implementation I did some further research and identified this exact limitation myself. I have a follow-up commit ready that addresses it by resolving the container's actual cgroup path via  `/proc/self/cgroup` before constructing the file paths.                                         What's your preference — should I push it as a follow-up commit on this PR, or would you rather it lands as a separate PR?

krunal1313:
@stafot You can add follow-up commit to this PR only.

stafot:
Done, @krunal1313

krunal1313:
@stafot 
Changes look good.
why have you modified onnxruntime.BUILD file?

stafot:
@krunal1313 I did some changes locally to be able to build on my local and probably accidentally committed, let me revert back this file.

krunal1313:
@stafot 
Could you please organize cgroup related resource files in better way?
Currently some are  in `test/resources`, some in `test/resources/cgroupsfs`

move all files in `test/resources/cgroup` in related sub-directories.

stafot:
done @krunal1313

stafot:
could any maintainer allow this to run https://github.com/typesense/typesense/actions/runs/26037528809 to check if it is safe for merge? cc @krunal1313 @kishorenc

nikoshet:
@krunal1313 any news on this?

stafot:
@krunal1313 could you reapprove test run to see if the fixtures fix did the trick?

kishorenc (approved):

Issue #2918, opened 2026-05-12
Index representing object[] values is desync with the real storage


rbtAtOnk:
## Bug Description
After multiple writes on an object[] some attributes are misindexed, notably after a PATCH which remove object in it : index keep ghost values when querying with filter_by. After a restart of the instance, the index is good. So the storage works as expected but not the tracking inside the associated index.

## Reproduction Steps
I have a 3 nodes cluster with a collection about 5 millions documents. The schema of this collection is above.

```bash
# Optional, generate dummy data
for i in $(seq 1 1000); do 
echo "{\"id\":\"$(uuidgen)\",\"title\":\"$(grep '^.\{5\}$' /usr/share/dict/words | shuf -n1) $(grep '^.\{5\}$' /usr/share/dict/words | shuf -n1)\",\"lastLease\":null,\"nextLease\":null,\"createdAt\":$(date -u '+%s'),\"updatedAt\":$(date -u '+%s'),\"users\":[$(shuf -i 2000-65000 -n 1), $(shuf -i 2000-65000 -n 1)],\"commentsToValidate\":[{\"userId\":$(shuf -i 2000-65000 -n 1),\"commentIds\":[$(shuf -i 2000-65000 -n 1),$(shuf -i 2000-65000 -n 1)]}],\"linkedBooks\":[{\"userId\": $(shuf -i 2000-65000 -n 1),\"bookIds\":[$(shuf -i 2000-65000 -n 1),$(shuf -i 2000-65000 -n 1)]}]}" >> import.jsonl
done

# Start a dummy server
docker run -p 8108:8108 \
  -v"$(pwd)"/data:/data typesense/typesense:30.1 \
  --data-dir /data \
  --api-key=dockerlocal \
  --enable-cors

# Export api key
export TYPESENSE_API_KEY=dockerlocal

# create the schema
curl -s "http://localhost:8108/collections" \
  -X POST \
  -H "Content-Type: application/json" \
  -H "X-TYPESENSE-API-KEY: $TYPESENSE_API_KEY" \
  -d "$(cat schema.json)"

# import dummy data
curl -s -H "X-TYPESENSE-API-KEY: $TYPESENSE_API_KEY" \
  -X POST \
  -T import.jsonl \
  "http://localhost:8108/collections/books/documents/import?action=create" > import.result.$(date -u '+%s').jsonl
grep -v true import.result.*.jsonl | wc -l

# select a random UUID in the import file
[...]

PR #2920, opened 2026-05-13
Fix stale nested object array indexes on patch update

krunal1313:
## Change Summary

PATCH updates that replaced a nested `object[]` field with an empty array could leave old flattened child field values in the in-memory index.

Example shape:

```json
{
  "reviewQueue": [
    { "reviewerId": 9001, "taskIds": [11, 22, 33] }
  ]
}
```

After this PATCH:

```json
{
  "reviewQueue": []
}
```

the stored document was correct, but a filter like:

```text
reviewQueue.reviewerId:=9001
```

could still return the document.

### Cause
Nested object arrays are indexed through flattened synthetic child fields such as:

```text
reviewQueue.reviewerId
reviewQueue.taskIds
```

During update diffing, the old parent field was added to `del_doc`, but the old flattened child fields were not always added. As a result, `Index::remove()` did not remove stale child field entries from the numeric/facet indexes.

### Fix

When a nested parent field is deleted or replaced, the update diff now checks the old document's `.flat` metadata and adds matching flattened child fields to `del_doc`.

So replacing `reviewQueue` also schedules removals for fields like:

```text
reviewQueue.reviewerId
reviewQueue.taskIds
```

This keeps search/filter indexes consistent with the stored document after PATCH updates.

## PR Checklist

### DIFF
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -8040,6 +8040,21 @@ void Index::get_doc_changes(const index_operation_t op, const tsl::htrie_map<cha
     nlohmann::json final_update_doc = nlohmann::json::object(); // Temporary for changed fields
 
     // Recursive function to apply merge patch and track changes
+    const auto add_nested_child_deletions = [](const std::string& key, const nlohmann::json& old,
+                                               nlohmann::json& del) {
+        if(!old.contains(".flat")) {
+            return;
+        }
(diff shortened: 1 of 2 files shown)


krunal1313:
Fixes #2918

PR #2934, opened 2026-05-29
fix: keep deferred collection state alive across drops

alexmac:
## Change Summary
Bug: deferred export/delete request state stored a raw Collection*. If a chunked export or delete paused, then the collection was dropped before the next chunk, the manager could destroy Collection -> Index -> num_tree_t while the request state still held a dangling pointer. This results in a use-after-free crash on the next deferred chunk when stateful_export_docs or stateful_remove_docs touches the freed Collection, or it could corrupt memory and later show up during Index / num_tree_t destruction like the stack trace.

Fix: export_state_t and deletion_state_t now hold std::shared_ptr<Collection> in core_api_utils.h (line 10), and the route handlers store the request’s shared pointer in core_api.cpp (line 1489) and core_api.cpp (line 2212). That keeps the collection alive until the deferred request state is destroyed.

## PR Checklist

### DIFF
--- a/include/core_api_utils.h
+++ b/include/core_api_utils.h
@@ -1,12 +1,13 @@
 #pragma once
 
 #include <cstdlib>
+#include <memory>
 #include <vector>
 #include "collection.h"
 #include "http_data.h"
 
 struct deletion_state_t: public req_state_t {
-    Collection* collection;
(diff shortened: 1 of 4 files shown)


alnr (approved):
This is a good change.

Issue #2861, opened 2026-04-01
Facet counts of joined fields


dejonghee:
This is a follow-up ticket for issue #1706 

This is a (simplified) example of one of our use cases:

Let's assume we have an e-commerce webshop where we sell t-shirts. We use a Product/Variant data model where the product is 'marketing' representation of the t-shirt which contains all common information and the variant is the actual SKU which only contains the specific info. Variants are defined by size (S/M/L) and color (Red/Blue/Green), but can also have a different price, stock level, etc.

On a product lister we want to show the product level to the customer (you don't want to see 3 t-shirts that look exactly the same, but have a different size). The product lister also has a facet map which contains a mix of Product (e.g. Brand) and Variant (e.g. Size & Color) facets. As we are showing the product level to the customer, we want the facet counts to reflect the number of products.

To implement this we are using a Product and a Variant index and we join the Product & Variant to calculate our facet map.

Suppose we only have 1 product (of Brand X) in our catalogue which is available in 8 variants (Red/S, Red/M, Red/L, Blue/S, Blue/M, Blue/L, Green/S, Green/M).

The facet map will look as follows:

Brand:
X (1)

Color:
Red (3)
Green (2)
Blue (3)

Size:
S (3)
M (3)
L (2)

The fact that the facet counts are a mix between Product (Brand) and Variants (Color & Size) counts is confusing for the end-user because on the product lister we only visualize the product level (so the customer only sees 1 product).

The expected facet map for our use case would be:

Brand:
X (1)

Color:
Red (1)
Green (1)
Blue (1)

Size:
S (1)
M (1)
L (1)

Would there be a way to achieve this outcome using the Typesense Join feature?

HTTP-file that can be used to reproduce the scenario:
```
@hostname = localhost:46675
@apiKey = typesense-api-key

### Start with a clean slate by deleting existing collections.

DELETE http://{{hostname}}/collections/products
X-TYPESENSE-API-KEY: {{ApiKey}}

DELETE http://{{hostname}}/collections/variants
X-TYPESENSE-API-KEY: {{ApiKey}}

### Create Products Collection.

POST http://{{hostname}}/collections
X-TYPESENSE-API-KEY: {{apiKey}}
Content-Type: application/json

{
  "name": "products",
  "fields": [
    {"name": "name", "type": "string", "facet": false, "sort": true},
    {"name": "brand", "type": "string", "facet": true, "sort": true}
  ],
[...]

Issue #2761, opened 2026-02-04
Faceting on joined reference fields gives unexpected results


forsdahl:
## Bug Description
Given a very simple example of products with brands and variants, faceting on reference fields gives unexpected results. Given that this is probably a very common use-case for faceting on reference fields, I think this case should be considered.
The problem is that given the scenario of products with variants, faceting on for example variant color should give facet counts from the queried collection (products in this case), not from the referenced collection (variants). Currently faceting on reference fields gives facet counts from the referenced collection, which in the use-case of a product listing with filters is wrong.
This problem and example is also highlighted in https://github.com/typesense/typesense/issues/1706
If the current behaviour is desired in some cases, perhaps some syntax for choosing which behaviour is wanted can be added?

## Reproduction Steps
```bash
set -x

export TYPESENSE_API_KEY=xyz
export TYPESENSE_HOST=http://localhost:8108

docker stop typesense-repro 2>/dev/null
docker rm typesense-repro 2>/dev/null
rm -rf "$(pwd)"/typesense-data-dir-repro
mkdir "$(pwd)"/typesense-data-dir-repro

# Wait for Typesense to be ready
docker run -d -p 8108:8108 --name typesense-repro \
            -v"$(pwd)"/typesense-data-dir-repro:/data \
            typesense/typesense:30.1 \
            --data-dir /data \
            --api-key=$TYPESENSE_API_KEY \
            --enable-cors

# Wait till typesense is ready.
until curl -s -o /dev/null -w "%{http_code}" "$TYPESENSE_HOST/health" -H "X-TYPESENSE-API-KEY: ${TYPESENSE_API_KEY}" | grep -q "200"; do
  sleep 2
done

curl -s "$TYPESENSE_HOST/debug" \
       -H "X-TYPESENSE-API-KEY: ${TYPESENSE_API_KEY}" | jq


curl -s "$TYPESENSE_HOST/collections" \
       -X POST \
       -H "Content-Type: application/json" \
       -H "X-TYPESENSE-API-KEY: ${TYPESENSE_API_KEY}" \
       -d '
          {
             "name": "Brands",
             "fields": [
               {"name": "name", "type": "string", "facet": true },
               {"name": "sort", "type": "int32" }
             ],
             "default_sorting_field": "sort"
           }' | jq

curl -s "$TYPESENSE_HOST/collections" \
       -X POST \
       -H "Content-Type: application/json" \
[...]

Issue #1706, opened 2024-05-03
[Feature Request] Add support for faceting on joined reference fields


mbombonato:
## Description

Consider the following schemas:
```
{
  "name":  "books",
  "fields": [
    {"name": "title", "type": "string"},
    {"name": "author_id", "type": "string", "reference": "authors.id"}
  ]
}

{
  "name":  "authors",
  "fields": [
    {"name": "id", "type": "string"},
    {"name": "name", "type": "string"}
  ]
}
```

I'd like to request a facet for the field "name" of the referenced collection "authors"
```
{
	"collection": "books",
	"include_fields": "$authors(name)",
	"q": "famous"
	"facet_by": "$authors(name)"
}
```

## Expected Behavior

Return the results with a facet for the field "authors.name"

## Actual Behavior

```
{
	"message": "Could not find a facet field named `$authors` in the schema."
}
```

## Metadata

**Typesense Version**: 26

**OS**: any

mbombonato:
I'd like to add the importance of this feature. I was about to change my data model to use the new JOIN feature of the Typesense 26. This would save us a huge amount of indexing time and data storage. Now I have to wait until this feature is released, because we need faceting working for referenced collections  :(

Adding support for JOINs was immense new feature, but still incomplete without the proper faceting support.

yanchogeorgiev:
Any plans to implement this soon? I think it's really important feature. We can't practically use joins without it.

vu3jej:
Stumbled upon this issue while trying to implement this functionality on the joins. Would love to see this feature in Typesense.

data-crowdanalytix:
Definitely important to us in our implementation

xian13:
Hi,

Since this issue hasn't appeared on the public roadmap yet, should we assume it will take a long time before it's implemented?

We are experiencing a similar issue in our implementation.

We have two collections:

product
price
We separate the price collection because a single product can have multiple prices based on member types. This also helps reduce indexing complexity, as updates to products and prices happen in different scenarios. Additionally, we want to minimize vector indexing in the product collection when we implement vector search later on, so price updates don't require reindexing the product.

We encountered this issue when trying to get the min, max, and average price values based on a query in the product collection.

An alternative solution could be querying the price collection and filtering by the product collection, but this feature is also not implemented and also not yet appeared on the public roadmap yet, as mentioned in #1657.

viktor-nikolaev:
yet another +1 for this feature

[...]

PR #2939, opened 2026-05-29
feat(facet): support :weighted modifier on reference facets

lBroth:
## Summary

Adds an opt-in `:weighted` modifier to `facet_by` for reference facets so each child reference weights the parent's value in `stats.sum` and counts (instead of the current default that collapses duplicates to one entry per parent).

```
facet_by=$other_collection(field):weighted
```

Default behavior is unchanged.

## Motivation

When a parent collection has many child references (cart items, folder items, collection duplicates), faceting on a parent field via a join currently sums the parent's value only once, regardless of how many child rows reference it. There is no built-in way to obtain a child-weighted aggregate from a single search.

Related discussion: #1706, #2761, #2861.

## Behavior

| Query | Default | `:weighted` |
|-------|---------|------------|
| 3 child rows referencing parent A (value=1000) + 1 child row referencing parent B (value=500) | sum=1500, count(1000)=1, count(500)=1 | sum=3500, count(1000)=3, count(500)=1 |

## Implementation

- `Index::get_reference_facet_ids` accepts `bool keep_child_multiplicity` and skips the final `erase(unique(...))` on `ref_doc_ids` when true.
- `Index::get_related_ids` / `get_related_ids_with_lock` accept a matching `bool keep_multiplicity` and skip their internal dedup so duplicates accumulate correctly across child docs.
- `use_value_index` is forced to `false` for weighted facets (the value-index intersection path collapses duplicates, defeating the multiplicity).
- `StringUtils::split_facet` was dropping any suffix after `)` for `$ref` facets; it now forwards an optional `:modifier` so it reaches `Collection::parse_facet`, which rejects unknown modifiers.
- The `weighted` bool is propagated through every `emplace_back` that rebuilds facets for batching (top_k, value, thread batches), so the flag survives the parallel facet path.

## Test plan

- [x] New regression test `CollectionJoinTest.FacetByReferenceWeightedAggregation` covering:
  - default dedup behavior (sum=1500)
  - weighted multiplicity (sum=3500, count distribution)
  - unknown-modifier rejection
- [x] `CollectionJoinTest.FacetBy*` — 3/3 pass (existing + new)
- [x] `CollectionFacetingTest.*` — 49/49 pass (incl. `FacetByReferenceWithJoinFilter`)
- [x] `FacetIndexTest.*` + `StringUtilsTest.*` — 21/21 pass
- [x] Manual JS harness reproducing the original use case (folder with duplicate items) sees `sum=1000*3 + 500=3500` with `:weighted`, `sum=1500` without

Total: 74 tests passing locally on a Linux arm64 build, zero regressions on the relevant suites.

This PR is opened as **draft** to validate CI before requesting review.

### DIFF
--- a/include/collection.h
+++ b/include/collection.h
@@ -1278,7 +1278,7 @@ public:
     Option<std::string> get_referenced_in_field_with_lock(const std::string& collection_name) const;
 
     Option<bool> get_related_ids_with_lock(const std::string& field_name, const std::vector<uint32_t>& seq_id_vec,
-                                           std::vector<uint32_t>& result) const;
+                                           std::vector<uint32_t>& result, bool keep_multiplicity = false) const;
 
     Option<bool> update_async_references_with_lock(const std::string& ref_coll_name, const std::string& filter,
                                                    const std::set<std::string>& filter_values,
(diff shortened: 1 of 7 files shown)


lBroth:
Closing — keeping work on personal fork for now, not requesting upstream review yet.

Issue #2827, opened 2026-03-11
async_reference JOINs with filter_by break after collection alias swap


cyppe:
## Summary

When using `async_reference: true` on a reference field that targets a **collection alias**, swapping that alias to a new collection breaks `filter_by` JOINs with:

```
Failed to join on `<new_collection_name>`: No reference field found.
```

This happens regardless of whether the old collection is deleted or still exists. The reference metadata appears to be permanently bound to the original concrete collection name at creation time, and is not updated when the alias target changes.

Notably, `include_fields` JOINs continue to work after the alias swap — only `filter_by` JOINs break.

## Context

This is a **separate issue** from #2680, which was about dangling async references after direct collection delete+recreate (fixed by PR #2682 in v30.1). The fix in #2682 works correctly for the non-alias case.

This issue specifically affects the **documented alias swap pattern** for zero-downtime reindexing ([Collection Alias docs](https://typesense.org/docs/guide/collection-alias.html)), which is the recommended way to reindex data without downtime.

A user has also reported what appears to be this same issue on v30.2.rc5, where `filter_by` with `$bookings(enterpriseId:='11')` returns `"No reference field found"` when using aliased collections.

## Reproduction

**Tested on:** `typesense/typesense:30.1`

### Minimal steps

1. Create `bookings_v1` collection
2. Create alias `bookings` → `bookings_v1`
3. Create `payments` collection with `"reference": "bookings.id", "async_reference": true`
4. Index documents in both collections
5. Verify `filter_by=$bookings(enterpriseId:=ent-1)` works (**it does**)
6. Create `bookings_v2` collection, populate it
7. Swap alias: `bookings` → `bookings_v2`
8. Same query now fails: `"Failed to join on bookings_v2: No reference field found."`

### Complete reproducible script

Save as `repro-alias-async-ref.sh` and run with `bash repro-alias-async-ref.sh`:

```bash
#!/bin/bash
# Reproduction script for: async_reference JOINs break after alias swap
# Tested on: typesense/typesense:30.1

### Configuration ########################################
export TYPESENSE_API_KEY=xyz
export TYPESENSE_HOST=http://localhost:8208
TYPESENSE_IMAGE="typesense/typesense:30.1"

PASS=0
FAIL=0

check_result() {
    local test_name="$1"
    local result="$2"
    local expect_success="$3"

    if [ "$expect_success" = "ok" ]; then
        if echo "$result" | grep -qi "error\|failed\|No reference field found"; then
[...]

PR #2963, opened 2026-06-26
Rebind reference field to the new swapped collection alias.

happy-san:
## Change Summary
Harden async reference handling when aliases are created, resolved, or retargeted. Introduce staged async-reference backfills so helper-field updates can be validated before metadata is mutated, then applied with rollback support if any child collection update fails.
  
  - Rebind alias-based reference fields when an alias target is swapped, including future-target aliases.
  - Resolve deferred alias references when the target collection is later created.
  - Stage async reference helper backfills before applying them, so partial failures can be rolled back.
  - Preserve collection lifetime during staged backfills by retaining shared_ptr<Collection> handles.
  - Apply helper-field-only patches against the current child document to avoid overwriting concurrent document updates.
  - Add regression coverage for alias target swap rollback, create-time deferred reference resolution, stale staged backfills, and dropped referencing collections during backfill.

## PR Checklist

### DIFF
--- a/include/collection.h
+++ b/include/collection.h
@@ -525,6 +525,7 @@ private:
 
     std::string get_seq_id_key(uint32_t seq_id) const;
 
+public:
     struct async_reference_backfill_update_t {
         uint32_t seq_id;
         nlohmann::json old_doc;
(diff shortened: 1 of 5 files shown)


kishorenc (approved):

alangmartini:
Fixed #2827

PR #2971, opened 2026-07-02
fix: validate user_id for log analytics events

tharropoulos:
## Change Summary
- reject log rules missing `user_id` with a 400 instead of hanging
- keep counter rules accepting missing `user_id` with empty user id

## PR Checklist


kishorenc (changes requested):
1. We have to also account for async handlers, which will hang when the new catch executes because routes such as search and multi-search have `async_res = true`. After catching an exception, the code sets a 500 response but skips STREAM_RESPONSE_MESSAGE, so the response is never dispatched. We should track a local should_dispatch flag and force it true in the catch, similar to the existing BatchedIndexer exception path.

  2. We should log e.what() but not expose it in API response because it can return implementation details. So let's log e.what(), but return a fixed error message.

PR #2984, opened 2026-07-15
feat: parallel HTTP event loops sized by --thread-pool-size

alnr:
## Change Summary

Run min(thread-pool-size, cores) h2o event loops instead of one. Each loop
has its own SO_REUSEPORT listener (the kernel load-balances accepts), h2o
context, message dispatcher and SSL refresh; loop 0 runs on the run()
thread. A connection is owned by exactly one loop: it is accepted, read
and written only there.

Every request is tagged with its owning loop's dispatcher
(http_req::res_dispatcher), and all response/request-proceed/deferral
messages are delivered through it via the new helpers
HttpServer::deliver_stream_response / deliver_request_proceed /
deliver_defer_processing. For requests with no owning loop (writes
replayed from the log, tests) the helpers mirror the dead-request paths
of the corresponding message handlers.

The global dispatcher access paths are deleted -- HttpServer::
get_message_dispatcher(), the unused HttpServer::send_message() wrapper
and ReplicationState's stashed message_dispatcher -- so no delivery site
can bypass the owning loop. That makes multiple loops safe in Raft mode
too: the BatchedIndexer apply path, follower->leader forwarding
(HttpClient), the SSE proxy and all write-rejection paths now deliver on
the request's own loop.

Deferred-processing timers are linked on the owning loop (the dispatcher
carries its h2o loop), so streamed responses (e.g. /documents/export) no
longer stall under backpressure when the request belongs to a loop other
than loop 0.

Removes the single-event-loop ceiling on HTTP intake and response I/O.

Three adjustments follow from requests now being minted on N threads:
http_req::start_ts, which doubles as the unique write-request ID in the
BatchedIndexer (request map and write-log chunk keys), comes from a global
atomic monotonic counter since wall-clock microseconds are no longer unique
across accept threads; deliver_defer_processing drives dispatcher-less
(replayed) write continuations directly on the worker pool, so chunked
handlers such as batched deletes still apply fully on followers and during
log replay; and the access log's ofstream is guarded by a mutex now that
requests are logged from multiple loops.

## What is now parallel that wasn't before?

```
| Work                                                                                                               | Where                                                                                       | CPU significance                           |
| ------------------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------- | ------------------------------------------ |
| TCP accept                                                                                                         | `HttpServer::on_accept`, `src/http_server.cpp`                                              | negligible                                 |
| TLS handshake (asymmetric crypto)                                                                                  | h2o + per-loop `accept_ctx.ssl_ctx`                                                         | **high** per new connection                |
| TLS record encrypt/decrypt of all request/response bytes (AES-GCM)                                                 | h2o socket layer on the owning loop                                                         | **high**, proportional to bytes            |
| HTTP/1 parsing, HTTP/2 framing + HPACK                                                                             | h2o internals, in-loop                                                                      | moderate                                   |
| Routing, query-string parse, url-decode, CORS, header extraction                                                   | `catch_all_handler`                                                                         | minor                                      |
| Authentication incl. HMAC-SHA256 for scoped API keys (regular + multi_search)                                      | `catch_all_handler` / `process_request` → `AuthManager::authenticate` → `StringUtils::hmac` | **high** with scoped keys                  |
| Access-log key-prefix HMAC + log write                                                                             | `catch_all_handler` → `AppMetrics::write_access_log`                                        | moderate (if enabled)                      |
| Rate-limit check                                                                                                   | `RateLimitManager::is_rate_limited`                                                         | minor (but see locks below)                |
| Request-body memcpy (×2 for buffered: h2o entity → local, local → `http_req`)                                      | `catch_all_handler:604` + `http_req` ctor                                                   | **high** for large bodies                  |
| Streaming-import body chunk aggregation (`body += chunk`)                                                          | `async_req_cb`                                                                              | **high** for imports                       |
| Write path, pre-Raft: resource check, route lookup, alter check                                                    | `ReplicationState::write`                                                                   | minor                                      |
| Write path: gzip **inflate** of compressed request bodies                                                          | `handle_gzip` → `inflate`, `src/raft_server.cpp:274`                                        | **high** for compressed imports            |
| Write path: raft-log entry serialization — `request->to_json()` = full-body copy + JSON dump (+ base64 for binary) | `src/raft_server.cpp:367`, `include/http_data.h:481`                                        | **high** — real serialization on the loop  |
| Response gzip **compression** (h2o compress filter, min 256 B, quality 1)                                          | `h2o_send` path; registered `src/http_server.cpp:370`                                       | **high** for large search/export responses |
| Response delivery: `h2o_start_response`/`h2o_send`, generator proceed, defer-timer scheduling                      | `on_stream_response_message`, `defer_processing`                                            | minor per chunk                            |
| Per-loop SSL cert refresh (disk read + `SSL_CTX` rebuild)                                                          | `on_ssl_refresh_timeout`                                                                    | rare latency blip, per loop                |
```

## PR Checklist

### DIFF
--- a/include/app_metrics.h
+++ b/include/app_metrics.h
@@ -26,6 +26,9 @@ private:
     std::string access_log_path;
     std::ofstream access_log;
 
+    // guards `access_log`: written from every HTTP event loop, flushed from a timer on loop 0
+    std::mutex access_log_mutex;
+
     AppMetrics() {
         current_counts = new spp::sparse_hash_map<std::string, uint64_t>();
         counts = new spp::sparse_hash_map<std::string, uint64_t>();
--- a/include/http_data.h
+++ b/include/http_data.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <algorithm>
+#include <atomic>
 #include <cstdint>
 #include <string>
 #include <map>
--- a/include/http_data.h
+++ b/include/http_data.h
@@ -256,12 +256,20 @@ public:
     }
 };
 
+struct http_message_dispatcher;
+
 struct http_req {
     static constexpr const char* AUTH_HEADER = "x-typesense-api-key";
     static constexpr const char* USER_HEADER = "x-typesense-user-id";
     static constexpr const char* AGENT_HEADER = "user-agent";
     static constexpr const char* CONTENT_TYPE_HEADER = "content-type";
     static constexpr const char* OCTET_STREAM_HEADER_VALUE = "application/octet-stream";
+
+    // The message dispatcher of the event loop that owns this request's connection. Responses and
+    // deferred work MUST be delivered through this dispatcher, since only the owning loop may touch
+    // the connection's socket. nullptr means the request never belonged to an event loop
+    // (e.g. a write replayed from the log, or tests).
+    http_message_dispatcher* res_dispatcher = nullptr;
 
     h2o_req_t* _req;
     std::string http_method;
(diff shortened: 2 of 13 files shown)


alnr:
A follow-up commit replaces the per-loop SO_REUSEPORT listeners with the model
h2o's own server uses by default (`dup_listener`, h2o `src/main.c`): one bound
socket, each loop polling its own `dup()` of it. The loops race non-blocking
accepts on the shared queue; losing the race is a no-op (EAGAIN), and the
level-triggered evloop keeps firing while connections remain queued.

Rationale: a second process binding the same port fails with EADDRINUSE again
(SO_REUSEPORT let it bind and silently steal connections); the shared queue is
pull-based, so idle loops win the race and load follows capacity, whereas
SO_REUSEPORT hash-pins connections to a loop regardless of its backlog; and
plain SO_REUSEPORT does not load-balance TCP on macOS/BSDs (h2o only enables
reuseport on Linux/FreeBSD).

Each loop accepts one connection per wakeup — deliberately not h2o's 10-accept
batch — trading a few extra wakeups for finer spreading of connections across
loops. Also fixes bind/dup error reporting: errno is now captured before
LOG(), whose constructor can clobber it (bind failures logged ENOENT instead
of EADDRINUSE).

PR #3010, opened 2026-08-02
Fix out-of-bounds read when a query has multiple `_eval` sort clauses

### LINKED ISSUE #3009: Multiple `_eval` clauses in `sort_by` share one cursor vector (out-of-bounds read)
## Bug Description

When a query has **more than one `_eval` clause** in `sort_by`, all of them share a single cursor
vector inside `Index::compute_sort_scores`. That vector is sized once, from whichever clause is
scored first. A later clause with more expressions then indexes past the end of it.

Two effects:

- **Out-of-bounds heap read** when the second clause has more expressions than the first.
- **Cross-clause cursor corruption** even when the sizes match, because the cursor is forward-only
  and every clause advances the same one. Clause B's cursor gets advanced by clause A's scan, so
  clause B can miss matches it should have found.

In `src/index.cpp`, `compute_sort_scores`:

```cpp
if (filter_indexes.empty()) {
    filter_indexes = std::vector<uint32_t>(count, 0);   // sized by the FIRST clause only
}

for (; eval_index < count; eval_index++) {
    ...
    auto& filter_index = filter_indexes[eval_index];    // indexed by THIS clause's count
```

`filter_indexes` is a single `std::vector<uint32_t>` passed by reference from the caller (declared
as `eval_filter_indexes` at each call site) and reused for every `_eval` sort field in the query.

## Reproduction Steps

```bash
export TS="http://localhost:8108"
export KEY="xyz"

curl -s -X DELETE "$TS/collections/eval_cursors" -H "X-TYPESENSE-API-KEY: $KEY" > /dev/null

curl -s -X POST "$TS/collections" \
  -H "X-TYPESENSE-API-KEY: $KEY" -H "Content-Type: application/json" -d '{
    "name": "eval_cursors",
    "fields": [
      {"name": "domain", "type": "string"},
      {"name": "skills", "type": "string[]"}
    ]
  }'

curl -s -X POST "$TS/collections/eval_cursors/documents/import?action=create" \
  -H "X-TYPESENSE-API-KEY: $KEY" -H "Content-Type: text/plain" --data-binary '
{"id":"1","domain":"engineering","skills":["graphql"]}
{"id":"2","domain":"engineering","skills":["react"]}
{"id":"3","domain":"design","skills":["typescript"]}'

# First _eval clause has 1 expression, second has 3.
curl -s -G "$TS/collections/eval_cursors/documents/search" \
  -H "X-TYPESENSE-API-KEY: $KEY" \
  --data-urlencode 'q=*' \
  --data-urlencode 'sort_by=_eval([(domain:=engineering):5]):desc, _eval([(skills:=react):1,(skills:=typescript):2,(skills:=graphql):3]):desc'
```

The second clause reads `filter_indexes[1]` and `filter_indexes[2]` from a vector that was allocated
with size 1.

I also extracted the loop above verbatim into a standalone harness with the same two clauses and ran
it under AddressSanitizer, which reports:

```
==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x502000000134
READ of size 4 at 0x502000000134 thread T0
    #0 score_doc_old ... :21
0x502000000134 is located 0 bytes after 4-byte region [0x502000000130,0x502000000134)
```

(The ASAN run is of the extracted loop, not of a full `typesense-server` build — I don't have an
ASAN server build to hand. The out-of-bounds index is visible directly in the source above.)

## Expected vs Actual

**Expected behavior**
Documents ordered by the first clause, ties broken by the second:

| id | tier 1 | tier 2 |
|---|---|---|
| 1 | 5 (engineering) | 3 (graphql) |
| 2 | 5 (engineering) | 1 (react) |
| 3 | 0 | 2 (typescript) |

→ `1, 2, 3`

**Actual behavior**
The second clause reads uninitialised memory past the end of its cursor vector, so its cursors are
garbage. Ordering is undefined — it may return the right answer, the wrong order, or crash,
depending on what happens to sit past the allocation. This is undefined behaviour rather than a
deterministic wrong answer.

## Environment

- **Typesense version:** current `v31` branch (`be2c5cc1`). I haven't checked other branches.
- **Operating system:** Ubuntu x86_64

## Schema / Configuration

```json
{
  "name": "eval_cursors",
  "fields": [
    { "name": "domain", "type": "string" },
    { "name": "skills", "type": "string[]" }
  ]
}
```

## Suggested fix

Key the cursors by sort-field index instead of sharing one flat vector, so each `_eval` clause keeps
its own state. The reference-sort branch needs its cursors rewound per document (it already
reassigns the vector every document today), while the normal branch needs them preserved across
documents for the forward-only scan — so the two want distinct accessors.

Happy to send a PR for this if useful.

MutharasuArchunan13:
Fixes #3009.

> Earlier revisions of this PR described the trigger as two plain `_eval` clauses. That was wrong —
> validation rejects those with *"Only one sorting eval expression is allowed"*, and the regression
> test I first pushed failed in CI for exactly that reason. The real trigger and a working
> reproduction are below.

## Problem

`compute_sort_scores` shares one `filter_indexes` cursor vector across every `_eval` sort field, and
sizes it from whichever clause is scored first:

```cpp
if (filter_indexes.empty()) {
    filter_indexes = std::vector<uint32_t>(count, 0);   // sized by the first clause
}
...
auto& filter_index = filter_indexes[eval_index];        // indexed by this clause's count
```

Two `_eval` clauses *can* coexist, despite the `eval_sort_count` guard: that counter is only
incremented in the non-reference branch, because the `$Collection(...)` branch `continue`s before
reaching it. So a **reference `_eval` plus a normal `_eval`** passes validation, and both land in the
eval branch of `compute_sort_scores`.

From there:

- the reference clause reassigns the shared vector to `ref_seq_ids.size()` for every document;
- the normal clause then sees it non-empty, skips its own sizing, and indexes it with its own
  expression count.

When the referenced-doc count is smaller than the expression count, the reads run past the end of the
allocation. `lower_bound` starts from whatever happens to be on the heap, so matches get missed — and
missed differently on each request.

## Reproduction (released v29.0, no build needed)

Full script in #3009. Every document is in stock so the reference clause ties and the 16-expression
clause alone decides the order; document `i` matches only `f(i+1)`, worth `16-i`, so the correct
order is `0,1,...,15`:

```
run1 buggy : ['0','1','3','4','5','7','11','15','14','13','12','10','9','8','6','2']
run2 buggy : ['0','1','3','4','5','7','8','9','11','12','13','15','14','10','6','2']
run3 buggy : ['0','1','3','12','13','15','14','11','10','9','8','7','6','5','4','2']
control    : ['0','1','2','3','4','5','6','7','8','9','10','11','12','13','14','15']
```

Three identical requests, three different orders. The control — the same `_eval` without the
reference clause — is stable and correct.

It only bites when the first expression fails to match, since first-match semantics `break` before
reaching the out-of-range indices. That is why a short expression list usually looks fine.

## Fix

`eval_filter_indexes_t` (`include/index.h`) keys the cursors by sort-field index so each clause keeps
its own state. Two accessors, because the call sites need opposite lifetimes:

- `get(sort_field_index, expression_count)` — allocated on first use and **preserved across
  documents**, which the forward-only scan depends on.
- `reset(sort_field_index, count)` — **rewound every document**, matching what the reference branch
  already did (it reassigned the whole vector per document) and still needs, since the referenced doc
  set varies per document.

Scoring logic is untouched; this is only about where the cursors live.

## Test

`CollectionSortingTest.ReferenceAndNormalEvalKeepSeparateCursors` builds the scenario above —
referenced-doc count of 1 against 16 expressions — and asserts the full `0..15` ordering.

## Note

While here I noticed the reference branch reuses a single cursor (`ref_filter_indexes[j]`, keyed by
referenced-doc index) across all eval expressions in its inner loop, so `lower_bound` for expression
*n+1* starts from a cursor left by expression *n* over a different array. That looks unintended, but
it is pre-existing behaviour with existing coverage, so I have left those semantics exactly as they
were rather than changing them in a bugfix. Happy to file it separately.

### DIFF
--- a/include/index.h
+++ b/include/index.h
@@ -398,6 +398,40 @@ struct group_by_field_it_t {
     bool is_array;
 };
 
+/// Forward-only cursors into each `_eval` expression's sorted seq_id array, held per sort field.
+///
+/// `compute_sort_scores` is called with monotonically increasing seq_ids within a query, so each
+/// expression only ever needs to scan forward. The cursors must be kept separate per sort field:
+/// a query may carry more than one `_eval` clause, and sharing a single flat vector between them
+/// both corrupts their cursors and lets a clause with more expressions index out of bounds.
+struct eval_filter_indexes_t {
+    std::vector<std::vector<uint32_t>> per_sort_field;
+
+    /// Cursors for `sort_field_index`, allocated on first use and preserved across documents.
+    std::vector<uint32_t>& get(const size_t sort_field_index, const size_t expression_count) {
+        auto& indexes = slot(sort_field_index);
+        if (indexes.size() != expression_count) {
+            indexes.assign(expression_count, 0);
+        }
+        return indexes;
+    }
+
+    /// Cursors for a reference sort, rewound every document because the referenced doc set varies.
+    std::vector<uint32_t>& reset(const size_t sort_field_index, const size_t count) {
+        auto& indexes = slot(sort_field_index);
+        indexes.assign(count, 0);
+        return indexes;
+    }
+
+private:
+    std::vector<uint32_t>& slot(const size_t sort_field_index) {
+        if (per_sort_field.size() <= sort_field_index) {
+            per_sort_field.resize(sort_field_index + 1);
+        }
+        return per_sort_field[sort_field_index];
+    }
+};
+
 #ifdef TEST_BUILD
     extern bool testing_not_equals_bug;
 #endif
@@ -1170,7 +1204,7 @@ public:
                                      std::array<spp::sparse_hash_map<uint32_t, int64_t, Hasher32>*, 3> field_values,
                                      const std::vector<size_t>& geopoint_indices, uint32_t seq_id,
                                      const std::map<basic_string<char>, reference_filter_result_t>& references,
-                                     std::vector<uint32_t>& filter_indexes, int64_t max_field_match_score,
+                                     eval_filter_indexes_t& filter_indexes, int64_t max_field_match_score,
                                      int64_t* scores, int64_t& match_score_index,
                                      float vector_distance = 0) const;
 
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -3785,7 +3785,7 @@ Option<bool> Index::search(std::vector<query_tokens_t>& field_query_tokens, cons
             search_cutoff = search_cutoff || filter_result_iterator_no_groups->validity == filter_result_iterator_t::timed_out;
 
             std::vector<uint32_t> nearest_ids;
-            std::vector<uint32_t> eval_filter_indexes;
+            eval_filter_indexes_t eval_filter_indexes;
 
             std::vector<group_by_field_it_t> group_by_field_it_vec;
             if (group_limit != 0) {
@@ -4239,7 +4239,7 @@ Option<bool> Index::search(std::vector<query_tokens_t>& field_query_tokens, cons
             }
 
             std::vector<uint32_t> vec_search_ids;  // list of IDs found only in vector search
-            std::vector<uint32_t> eval_filter_indexes;
+            eval_filter_indexes_t eval_filter_indexes;
 
             std::vector<group_by_field_it_t> group_by_field_it_vec;
             if (group_limit != 0) {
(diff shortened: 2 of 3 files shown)


a-mutharasu:
@jasonbosco @kishorenc could you guys check this out.

kishorenc (approved):

PR #3011, opened 2026-08-02
Add a `sum` mode to `_eval` so multiple matching filter scores add up

### LINKED ISSUE #2014: [Feature] Allow _eval to sum the filter scores
## Description

I want to perform a search query with boost on two fields, following the logic:

A match on both fields should have the highest score.

A match on the first field should have a higher score compared to the second field.

The documentation is somewhat unclear regarding how sorting based on filter scores works when a document matches multiple filter expressions.

In my search query, I am observing an unexpected document order with 
`_eval([(bar:=2):2,(foo:=2):1]):desc,order:asc` 
It seems like the documents are being sorted by the maximum filter score instead of the sum of filter scores.

I have found a workaround by adding an additional filter:
`_eval([(bar:=2 && foo:=2):3,(bar:=2):2,(foo:=2):1]):desc,order:asc`

While this solution works for just two fields, it leads to a combinatorial explosion if more fields need to be handled. Using a formula like `2^n - 1`, the number of filters grows rapidly:

- 2 fields = 3 filters
- 3 fields = 7 filters
- 4 fields = 15 filters

I am not entirely sure, but this behavior might be intended, although it seems inefficient.

## Steps to reproduce

1) Cleanup

```http request
DELETE http://localhost:8108/collections/test
X-Typesense-Api-Key: testapikey
```

2) Create test collection

```http request
POST http://localhost:8108/collections
X-Typesense-Api-Key: testapikey
Content-Type: application/json

{
  "name": "test",
  "fields": [
    {"name": "foo", "type": "int32"},
    {"name": "bar", "type": "int32"},
    {"name": "order", "type": "int32","sort": true}
  ]
}
```

3) Add test documents

```http request
POST http://localhost:8108/collections/test/documents/import
X-Typesense-Api-Key: testapikey
Content-Type: application/jsonl

{"id": "1","foo": 1,"bar": 1, "order": 1}
{"id": "2","foo": 1,"bar": 2, "order": 2}
{"id": "3","foo": 2,"bar": 1, "order": 3}
{"id": "4","foo": 2,"bar": 2, "order": 4}
```

4) Perform search request with two filter scores

```http request
POST http://localhost:8108/multi_search
X-Typesense-Api-Key: testapikey
Content-Type: application/json

{
  "searches": [
    {
      "q": "*",
      "filter_by": "(foo:2 || bar:2)",
      "sort_by": "_eval([(bar:=2):2,(foo:=2):1]):desc,order:asc",
      "offset": 0,
      "limit": 10,
      "collection": "test"
    }
  ]
}
```

Actual result:

```json
{
  "results": [
    {
      "facet_counts": [],
      "found": 3,
      "hits": [
        {
          "document": {
            "bar": 2,
            "foo": 1,
            "id": "2",
            "order": 2
          },
          "highlight": {},
          "highlights": []
        },
        {
          "document": {
            "bar": 2,
            "foo": 2,
            "id": "4",
            "order": 4
          },
          "highlight": {},
          "highlights": []
        },
        {
          "document": {
            "bar": 1,
            "foo": 2,
            "id": "3",
            "order": 3
          },
          "highlight": {},
          "highlights": []
        }
      ],
      "out_of": 4,
      "page": 1,
      "request_params": {
        "collection_name": "test",
        "first_q": "*",
        "per_page": 10,
        "q": "*"
      },
      "search_cutoff": false,
      "search_time_ms": 0
    }
  ]
}
```

5) Perform search request with workaround

```http request
POST http://localhost:8108/multi_search
x-Typesense-Api-Key: testapikey
Content-Type: application/json

{
  "searches": [
    {
      "q": "*",
      "filter_by": "(foo:2 || bar:2)",
      "sort_by": "_eval([(bar:=2 && foo:=2):3,(bar:=2):2,(foo:=2):1]):desc,order:asc",
      "offset": 0,
      "limit": 10,
      "collection": "test"
    }
  ]
}
```

Actual result

```json
{
  "results": [
    {
      "facet_counts": [],
      "found": 3,
      "hits": [
        {
          "document": {
            "bar": 2,
            "foo": 2,
            "id": "4",
            "order": 4
          },
          "highlight": {},
          "highlights": []
        },
        {
          "document": {
            "bar": 2,
            "foo": 1,
            "id": "2",
            "order": 2
          },
          "highlight": {},
          "highlights": []
        },
        {
          "document": {
            "bar": 1,
            "foo": 2,
            "id": "3",
            "order": 3
          },
          "highlight": {},
          "highlights": []
        }
      ],
      "out_of": 4,
      "page": 1,
      "request_params": {
        "collection_name": "test",
        "first_q": "*",
        "per_page": 10,
        "q": "*"
      },
      "search_cutoff": false,
      "search_time_ms": 0
    }
  ]
}
```

## Expected Behavior

Documents should be ordered by the sum of filter scores, in the order: id=4,2,3.

## Actual Behavior

Documents are ordered by the maximum filter score, in the order: id=2,4,3

## Metadata:

**Typesense Version**: 27.1 running on Docker

**OS**: Ubuntu 24.04, running Docker 24.0.5

a-mutharasu:
Fixes #2014.

## Problem

`_eval([(a):1, (b):2, (c):4])` returns the score of the first matching expression and stops. A
`sort_by` written as a list of additive boosts therefore does not add: a document matching the last
three rules scores 4, not 7, and loses to one matching only the first. Nothing in the response
signals it, and reordering the expressions changes the result.

Expressing a real additive model meant enumerating every combination, which needs `2^n - 1`
expressions — 15 rules for 4 signals.

## Change

`mode: sum` adds the scores of every expression that matches, so `n` signals need `n` expressions:

```
sort_by=_eval([(domain:=engineering):25, (seniority:=3):6, (skills:=react):8], mode: sum):desc
sort_by=_eval((brand:nike), mode: sum):desc
```

| document | default | `mode: sum` |
|---|---|---|
| engineering, seniority 1 | 25 | 25 |
| design, seniority 3, react | 6 | **14** |

The default stays `first_match`. Summing by default would silently re-rank every released query and
break the if/else-if idiom, where a document that is both in stock and backordered should score as
in stock rather than as the sum of both.

## Parsing

The array form is delimited by `]`, so its parameters are unambiguous.

The single-expression form has no such delimiter, and a filter value may legally hold a top level
comma — `_eval(title:Hello, World)` matches that literal string today. Parameters are recognised
there only when the expression is wrapped in its own parentheses **and** the trailing span is
introduced by a comma. Everything else parses exactly as it does now:

```
_eval((brand:nike), mode: sum)           -> expression `brand:nike`, mode sum
_eval(brand:nike, mode: sum)             -> filter, unchanged
_eval(title:Hello, World)                -> filter, unchanged
_eval((a:1) && (b:2))                    -> filter, unchanged
_eval((brand:nike || brand:air) && x:1)  -> filter, unchanged
```

A recognised `mode` carrying an unusable value (`mode: sumx`) is an error rather than being left in
the expression, so a typo fails loudly instead of silently becoming a filter that matches nothing.

Reference `include_fields` parameters are now split with the same nesting awareness, so a nested
`sort_by` carrying a comma is no longer torn apart at it.

## Referenced collections

Supported. Each referenced document is scored on its own and the results are reduced to one value
for the source document: the maximum for a descending sort, the minimum for an ascending one, so the
reduction always favours the end of the ordering the caller asked for. A source document reaching
the same referenced document through several join paths scores it once. No referenced documents
scores 0.

## Overflow

Weights are accumulated in `__int128` and clamped once per document to `[INT64_MIN, INT64_MAX]`.
Accumulating exactly and clamping at the end keeps the total independent of the order the
expressions are listed in, which a per-addition saturating add does not.

The full int64 range stays usable: rather than negating the score for an ascending sort, which is
undefined for `INT64_MIN`, the ordering is reversed with `-x - 1` computed in `__int128`. That is a
bijection over int64, so no value has to be sacrificed to the clamp.

## Tests

- `EvalSumAddsEveryMatchingExpression` — the order inverts between the default and `mode: sum` on
  identical rules
- `EvalSumOnReferencedCollection` — per-referenced-document scoring and the max/min reduction
- `ReferencedEvalSumIsOrderIndependentNearOverflow`
- `EvalSumReferencedValidationFailureDoesNotLeak`
- `EvalSumSaturatesInsteadOfOverflowing`
- `ParseSortByClause` — both forms, the compound-filter cases above, and the invalid-mode error
- join tests covering the nested-comma `include_fields` split

## Credit

Commits 2-7 are @happy-san's: the validation-path leak fix, referenced-collection scoring, `__int128`
accumulation, the `include_fields` comma split, `INT64_MIN` preservation, and an `api_tests` fix that
consumes the import response so the search no longer races indexing.

The leak and the order-dependence of my original saturating add were both real defects in my commit.

### DIFF
--- a/src/collection_manager.cpp
+++ b/src/collection_manager.cpp
@@ -1770,9 +1770,51 @@ AuthManager& CollectionManager::getAuthManager() {
     return auth_manager;
 }
 
+/// Parses the optional trailing parameters of a multi-eval clause, i.e. everything between the
+/// closing `]` and the closing `)` of `_eval([...], <params>)`.
+static bool parse_multi_eval_params(const std::string& params_str, sort_by::eval_mode_t& eval_mode) {
+    auto params_substr = params_str;
+    StringUtils::trim(params_substr);
+    if (params_substr.empty()) {
+        return true;
+    }
+
+    // The span starts right after `]`, so a parameter list opens with the separating comma.
+    if (params_substr[0] != ',') {
+        return false;
+    }
+    params_substr = params_substr.substr(1);
+
+    std::vector<std::string> params;
+    StringUtils::split(params_substr, params, ",");
+    if (params.empty()) {
+        return false;
+    }
+
+    for (const auto& param: params) {
+        // `split` trims each token, so `mode:sum` and `mode: sum` both arrive here the same way.
+        std::vector<std::string> param_parts;
+        StringUtils::split(param, param_parts, ":");
+        if (param_parts.size() != 2 || param_parts[0] != sort_field_const::eval_mode) {
+            return false;
+        }
+
+        if (param_parts[1] == sort_field_const::eval_mode_sum) {
+            eval_mode = sort_by::eval_mode_t::sum_matches;
+        } else if (param_parts[1] == sort_field_const::eval_mode_first_match) {
+            eval_mode = sort_by::eval_mode_t::first_match;
+        } else {
+            return false;
+        }
+    }
+
+    return true;
+}
+
 bool parse_multi_eval(const std::string& sort_by_str, uint32_t& index, std::vector<sort_by>& sort_fields) {
     // FORMAT:
     // _eval([ (<expr_1>): <score_1>, (<expr_2>): <score_2> ]):<order>
+    // _eval([ (<expr_1>): <score_1>, (<expr_2>): <score_2> ], mode: sum):<order>
 
     std::vector<std::string> eval_expressions;
     std::vector<std::int64_t> scores;
@@ -1826,6 +1868,20 @@ bool parse_multi_eval(const std::string& sort_by_str, uint32_t& index, std::vect
         scores.emplace_back(std::stoll(score));
     }
 
+    // `index` is at the closing `]`, so everything up to the closing `)` of `_eval(` is the optional
+    // parameter list. It has to be consumed before the scan below, which stops at the first `:`.
+    // With a parameter present that first colon belongs to `mode: sum`, not to the order.
+    auto const close_paren_pos = sort_by_str.find(')', index);
+    if (close_paren_pos == std::string::npos) {
+        return false;
+    }
+
+    sort_by::eval_mode_t eval_mode = sort_by::eval_mode_t::first_match;
+    if (!parse_multi_eval_params(sort_by_str.substr(index + 1, close_paren_pos - index - 1), eval_mode)) {
+        return false;
+    }
+    index = close_paren_pos;
+
     while (++index < sort_by_str.size() && sort_by_str[index] != ':');
     if (index >= sort_by_str.size()) {
         return false;
@@ -1838,13 +1894,70 @@ bool parse_multi_eval(const std::string& sort_by_str, uint32_t& index, std::vect
     StringUtils::trim(order_str);
     StringUtils::toupper(order_str);
 
-    sort_fields.emplace_back(eval_expressions, scores, order_str);
+    sort_fields.emplace_back(eval_expressions, scores, order_str, eval_mode);
+    return true;
+}
+
+/// Splits a trailing parameter list off the single-expression form, `_eval((<expr>), mode: sum)`.
+///
+/// The array form is delimited by `]`, so everything after it is unambiguously a parameter. The
+/// single-expression form has no such delimiter and a filter value may legally hold a top level
+/// comma, as `title:Hello, World` does. Parameters are therefore only recognised when the
+/// expression is wrapped in its own parentheses, which lets a plain parenthesis count find where it
+/// ends. `_eval(brand:nike, mode: sum)` stays a filter, exactly as it parses today.
+///
+/// Returns false only when the parameter list is present but unusable.
+static bool split_trailing_eval_params(std::string& eval_expr, sort_by::eval_mode_t& eval_mode) {
+    auto expr = eval_expr;
+    StringUtils::trim(expr);
+    if (expr.empty() || expr[0] != '(') {
+        return true;
+    }
+
+    int paren_count = 0;
+    bool in_backtick = false;
+    size_t expr_end = std::string::npos;
+
+    for (size_t i = 0; i < expr.size(); i++) {
+        const char c = expr[i];
+        if (c == '`') {
+            in_backtick = !in_backtick;
+        } else if (in_backtick) {
+            continue;
+        } else if (c == '(') {
+            paren_count++;
+        } else if (c == ')' && --paren_count == 0) {
+            expr_end = i;
+            break;
+        }
+    }
+
+    if (expr_end == std::string::npos) {
+        return true;
+    }
+
+    // A parameter list is introduced by a comma. Anything else following the group belongs to the
+    // filter, as in `(a:1) && (b:2)`, and the expression is left exactly as it was.
+    auto trailing = expr.substr(expr_end + 1);
+    StringUtils::trim(trailing);
+    if (trailing.empty() || trailing[0] != ',') {
+        return true;
+    }
+
+    if (!parse_multi_eval_params(trailing, eval_mode)) {
+        return false;
+    }
+
+    // Drop the wrapping parentheses along with the parameter list.
+    eval_expr = expr.substr(1, expr_end - 1);
+    StringUtils::trim(eval_expr);
     return true;
 }
 
 bool parse_eval(const std::string& sort_by_str, uint32_t& index, std::vector<sort_by>& sort_fields) {
     // FORMAT:
     // _eval(<expr>):<order>
+    // _eval((<expr>), mode: sum):<order>
     std::string eval_expr = "(";
     int paren_count = 1;
     bool in_backtick = false;
@@ -1866,6 +1979,11 @@ bool parse_eval(const std::string& sort_by_str, uint32_t& index, std::vector<sor
         return false;
     }
 
+    sort_by::eval_mode_t eval_mode = sort_by::eval_mode_t::first_match;
+    if (!split_trailing_eval_params(eval_expr, eval_mode)) {
+        return false;
+    }
+
     while (sort_by_str[index] != ':' && ++index < sort_by_str.size());
     if (index >= sort_by_str.size()) {
         return false;
@@ -1880,7 +1998,7 @@ bool parse_eval(const std::string& sort_by_str, uint32_t& index, std::vector<sor
 
     std::vector<std::string> eval_expressions = {eval_expr};
     std::vector<int64_t> scores = {1};
-    sort_fields.emplace_back(eval_expressions, scores, order_str);
+    sort_fields.emplace_back(eval_expressions, scores, order_str, eval_mode);
 
     return true;
 }
--- a/api_tests/tests/nested_curly_filter.test.ts
+++ b/api_tests/tests/nested_curly_filter.test.ts
@@ -333,6 +333,15 @@ describe(Phases.SINGLE_FRESH, () => {
     });
     expect(res.ok).toBe(true);
 
+    const importResults = (await res.text())
+      .trim()
+      .split("\n")
+      .map((line) => JSON.parse(line) as { success: boolean });
+    expect(importResults).toHaveLength(items.length);
+    for (const result of importResults) {
+      expect(result.success).toBe(true);
+    }
+
     const twoPredicateExpected = countSameObjectMatches(
       items,
       (offer) => offer.quantities.some((quantity) => quantity >= 2) && offer.minCost <= 7500,
--- a/include/field.h
+++ b/include/field.h
@@ -501,6 +501,11 @@ namespace sort_field_const {
     static const std::string exclude_radius = "exclude_radius";
     static const std::string precision = "precision";
 
+    // `_eval([...], mode: <...>)` parameter and its accepted values.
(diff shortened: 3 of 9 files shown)


a-mutharasu:
Hi @jasonbosco @kishorenc, could you please take a look at this PR when you have a chance?

kishorenc:
I merged `https://github.com/typesense/typesense/pull/3010` but there is a conflict now. Can you please resolve?

a-mutharasu:
@kishorenc  Thanks for merging #3010  single commit now, conflict cleared. It was only a squash-merge SHA artifact, so nothing was hand-resolved. Also dropped the stale "stacked on" note from the description.

Still happy to switch `_eval_sum` to a parameter on `_eval` if you'd prefer that shape.

kishorenc:
I think we should make the `sum` a mode of eval instead of introducing `_eval_sum()`:

```
_eval([ (<expr_1>): <score_1>,..], mode: sum):<order>
```

The change required will be here in the [options parsing logic](https://github.com/typesense/typesense/blob/v31/src/collection_manager.cpp#L1828).

a-mutharasu:
Done — `mode` is now a parameter on `_eval`, and `_eval_sum()` is gone:

```
_eval([ (<expr_1>): <score_1>, .. ], mode: sum):<order>
```

`first_match` is the default, so existing queries are unchanged; `mode: first_match` is also accepted explicitly.

The one part worth a look is where you pointed: the order scan takes the **first** `:` after `]`, which with `mode: sum` present is `mode`'s colon rather than the one before `:desc`. So the span between `]` and the closing `)` is consumed as the parameter list first, then the existing scan runs unchanged.

Two things I defaulted on, easy to change:

- **Errors are generic.** A bad parameter gives *"Parameter `sort_by` is malformed."* rather than naming it, since `parse_multi_eval` returns `bool`. Happy to thread `Option<bool>` through for a specific message.
- **No `mode` on `_eval(<expr>)`.** Summing one expression equals first-match, so `parse_eval` is untouched.

One incidental tightening: `_eval([...]:desc` with no closing paren used to parse (the scan found a `:` either way) and is now rejected, since the parameter span is delimited by that paren. Nothing in the test suite relied on it.

a-mutharasu:
Pushed a follow up. While testing the parameter form I found a hole in the single expression form and fixed it.

`_eval(<expr>, mode: sum)` was being accepted with the parameter swallowed into the filter:

```
_eval(brand:nike, mode: sum):desc
  -> filter_expr = [brand:nike, mode: sum]   mode = first_match
```

`filter.cpp` has no top level comma separator, so that does not error. It just becomes a filter that matches nothing, and the boost silently does nothing. Same class of failure as the one this PR is about, so it seemed worth closing rather than documenting around.

It now takes the parameter properly. The tricky part is that this form has no `]` to delimit anything, and a top level comma can legally belong to the filter value:

```
_eval(title:Hello, World):desc
```

matches that literal string today. So the trailing span is only claimed when its first key names a known parameter, and everything else stays in the filter. Commas inside `[...]`, inside `(...)` and inside backticks are skipped, so `brand:[nike,adidas]`, `loc:(48.90,2.33,5.1 km)` and ``title:`Hello, mode: sum` `` all parse exactly as before.

One deliberate exception to that leniency: if the key is `mode` but the value is not usable, say `mode: sumx`, it errors instead of falling back to the filter. Otherwise a typo would go back to being a silent no op.

On the default, I kept `first_match` and I do not think it can be anything else. `_eval` is already released and documented, and summing by default would silently re-rank every existing query with two or more expressions where a document matches more than one. It would also break the if/else-if idiom outright, since a document that is both `in stock` and `backordered` would outscore one that is only `in stock`.

Tests cover both forms now, including the filter values with commas that have to be left alone.

a-mutharasu:
@kishorenc heads up on the red `api-tests` here. It is not coming from this PR, and I think it is worth a look rather than a retry.

The failure is `nested_curly_filter.test.ts:349`:

```
filter_by: offers.{quantities:>=2 && minCost:<=7500}
Expected: 4992
Received: 3662
```

**The input is fully deterministic.** `buildHighLoadItems()` seeds its generator with `rngFactory(42)`, so the 5000 documents are identical on every run, and `twoPredicateExpected` is computed in JS from those same documents. Both the data and the expected value are fixed at 4992.

**The server still returns different answers for it.** The earlier run on this branch returned 4992 and passed, this run returned 3662, with no code change in between that touches filtering. The same failure with the same two numbers is on `v31` itself at `2b1ea800`, so it is not specific to my branch.

Since `found` comes back short rather than wrong, my guess is that the search runs before the import has finished indexing every document, rather than the filter evaluating incorrectly. The test only asserts `res.ok` on the import and never inspects the per document results in the body, so a partially applied import would look exactly like this. Possibly related to the batched indexer work in 84ae2cfc, though I have not been able to confirm that.

I have deliberately left it out of this PR, since it is in the filter and indexing path while this change only touches `sort_by` parsing. Happy to look into it separately if that would help.

The `test` job, meaning the build and the C++ unit tests, is green.

src/collection_manager.cpp
```diff
@@ -1838,13 +1894,83 @@ bool parse_multi_eval(const std::string& sort_by_str, uint32_t& index, std::vect
...
+/// the expression, so that a typo such as `mode: sumx` fails loudly rather than turning into a
+/// filter that quietly matches nothing.
+static eval_params_result split_trailing_eval_params(std::string& eval_expr, sort_by::eval_mode_t& eval_mode) {
+    int paren_count = 0, bracket_count = 0;
+    bool in_backtick = false;
+    size_t split_pos = std::string::npos;
+
+    for (size_t i = 0; i < eval_expr.size(); i++) {
+        const char c = eval_expr[i];
+        if (c == '`' && (i == 0 || eval_expr[i - 1] != '\\')) {
```
happy-san: Hi @MutharasuArchunan13, We do not support escaping the ` backtick character. So we shouldn't introduce this logic here now.

happy-san: We should require the single expression to be included inside additional parentheses if `mode` option is to be provided.
```
_eval((brand:nike), mode: sum):desc
```
This filter expression can then be parsed with a simple parentheses count logic and the rest would be parsed as option.

a-mutharasu: Thanks @happy-san, both taken.

**Backtick escaping** — removed. You're right that nothing else in the parser treats `\` as an escape, so this was inventing a convention the engine doesn't have. It now matches the existing `parse_eval` handling: a backtick simply toggles the state.

**Wrapping parentheses for the parameter form** — adopted, and it made the code considerably smaller. The previous version only claimed a trailing span when its first key happened to be `mode`, purely so `_eval(title:Hello, World)` kept working. Requiring the parentheses makes that ambiguity structurally impossible, so the whole heuristic is gone along with its enum and its depth-tracked comma scan.

```
_eval((brand:nike), mode: sum):desc     -> expr=[brand:nike]  mode=sum
_eval(brand:nike, mode: sum):desc       -> filter, unchanged from today
_eval(title:Hello, World):desc          -> filter, unchanged from today
```

The array form is untouched, since `]` already delimits it.

One thing worth confirming is deliberate: `_eval((brand:nike), mode: sumx)` errors rather than falling back to treating the whole thing as a filter. Once the parentheses have identified a parameter list, a typo in it should fail loudly — falling back would restore exactly the silent no-op this PR exists to fix. Happy to change that if you'd rather it be lenient.

CI is running. The `api-tests` failure showing on earlier runs is unrelated to this PR: it is `nested_curly_filter.test.ts:349` asserting an exact `found` count on a time-bounded query, and it has failed and passed on identical commits here.

a-mutharasu: Correction to my previous reply, caught while testing: my first implementation of the parentheses rule broke compound filters.

`_eval((a:1) && (b:2))` starts with a parenthesised group, so a plain parenthesis count finds the end of that group and then treats ` && (b:2)` as the parameter list, which fails to parse. That would have rejected a filter that works today.

Fixed by only treating the trailing span as parameters when it is introduced by a comma. Anything else following the group belongs to the filter and the expression is left untouched.

```
_eval((brand:nike), mode: sum):desc            -> expr=[brand:nike]  mode=sum
_eval((a:1) && (b:2)):desc                     -> filter, unchanged
_eval((brand:nike || brand:air) && size:10)    -> filter, unchanged
```

Added tests covering both compound forms. Pushed as 22f8f492.

happy-san:
@MutharasuArchunan13 Could you rebase your PR to [this branch](https://github.com/happy-san/typesense/tree/pr_3011)?

src/collection_manager.cpp
a-mutharasu: updated

happy-san:
I have pushed one more commit. Let me know when you've rebased again.

happy-san:
@MutharasuArchunan13 Fixed the flaky test as well. The PR will be ready for @kishorenc's review after the rebase.

a-mutharasu:
Done #3011 now carries all six commits, at `7611a026`.

Thanks for building on it directly. All four additions are improvements on what I had:

- the validation-path leak was a real defect in my commit — the rejection ran after the recursive validation had already allocated filter trees
- accumulating in `__int128` and clamping once removes the order-dependence of my per-addition saturating add
- `reverse_eval_score` is a better answer than sacrificing `INT64_MIN` to the clamp, which is what I had done
- the `include_fields` comma splitting was a pre-existing bug that my syntax happened to surface

kishorenc (approved):

PR #3035, opened 2026-08-30
Fix union search returning fewer hits than `found`

alnr:
## Problem

A union search can silently drop documents: `found` is right, `hits` has fewer entries, `search_cutoff` is `false`.

The union topster identifies a document by `hash_combine(collection_id, seq_id)` (or `search_index, seq_id`). `hash_combine` is not injective for small integers: `(3, 1000)` and `(4, 932)` produce the same key, and so do any two documents of neighbouring collections whose sequence ids differ by roughly 64. The topster treats such a pair as one document and keeps only the higher-ranked one.

Repro: two collections with 100 documents each, union `q=*`, `per_page=250` → `found: 200`, 163 hits.

## Fix

Both parts of the key are 32-bit, so pack them (`id << 32 | seq_id`) instead of hashing them. The group-by path keeps `hash_combine`, as its second part is already a 64-bit group hash.

## Tests

- `TopsterTest.UnionKVKeyIsUniquePerDocument`: 32 ids × 4096 sequence ids give 131072 distinct keys in both modes.
- `UnionTest.ReturnsEveryHitAcrossCollections`: the repro above, asserting `found == hits == 200`.

### DIFF
--- a/include/topster.h
+++ b/include/topster.h
@@ -217,17 +217,21 @@ struct Union_KV : public KV {
                     std::tie(j->scores[0], j->scores[1], j->scores[2], i->search_index, j->key);
     }
 
+    // The key identifies one document of one search or collection. Both parts are 32-bit, so
+    // packing them is exact, whereas hashing them together collides between documents whose
+    // sequence ids differ by a small multiple of 64 (e.g. (3, 1000) and (4, 932)) and makes the
+    // topster drop one of them as a duplicate.
     static constexpr uint64_t get_key(const Union_KV* union_kv) {
         if(union_kv->remove_duplicates) {
-            return StringUtils::hash_combine(union_kv->collection_id, union_kv->distinct_key);
+            return (uint64_t(union_kv->collection_id) << 32) | union_kv->distinct_key;
         }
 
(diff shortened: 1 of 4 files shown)


kishorenc:
@alnr 

Thanks, this looks good. There is another place where we are doing the same thing and it will be good to fix that also.

The `hash_combine(collection_id, distinct_key)` is still used for populating `curated_union_keys`. With the default `remove_duplicates=true`, a curated `(3, 1000)` can therefore cause an unrelated raw `(4, 932)` hit to be skipped. Let's reuse `Union_KV::get_key(kv)` for both insertion and lookup, and add a curated/raw collision regression test as well.

alnr:
Good catch — done in 8da4ef75: both the `curated_union_keys` insertion and the raw-hit lookup now go through `Union_KV::get_key`, and `UnionTest.CuratedHitShouldNotSuppressUnrelatedRawHit` pins it (two collections × 100 docs, one curated include; under the old keys the curated `(c, 99)` collides with a raw document of the neighbouring collection for any collection id except 31, dropping it from the hits).

kishorenc (approved):

PR #3038, opened 2026-09-06
Add a gemma tokenizer type for EmbeddingGemma models

honzasterba:
Adds a `gemma` tokenizer type so `google/embeddinggemma-300m` can be used as a built-in model.

## Why a new tokenizer is needed

EmbeddingGemma tokenizes to raw SentencePiece ids wrapped in BOS=2 and EOS=1. Neither existing SentencePiece path produces those ids:

| Path | What it does | Why it does not fit |
|---|---|---|
| `xlm_roberta` | adds the fairseq `+1` offset to every id, `<s>`=0 / `</s>`=2 | every token id comes out shifted by one |
| `siglip` | lowercases, no BOS, EOS=1, caps at 64 tokens | case is destroyed, sequence is wrong |

This is not a small quality loss. Feeding EmbeddingGemma through the `xlm_roberta` path and comparing against `sentence-transformers`:

| Text | cosine |
|---|---|
| "The Eiffel Tower is located in Paris, France." | 0.0556 |
| "Python is a high-level programming language." | 0.2205 |
| "Der Hund läuft schnell durch den Park." | 0.1116 |

Those are unrelated vectors, so the model cannot be supported without its own tokenizer.

## The change

`GemmaTokenizer` is a plain SentencePiece tokenizer — raw ids, no lowercasing, BOS and EOS added, truncated to the model's 2048-token limit with EOS kept last. It mirrors `SigLIPTokenizer` in shape. Four call sites: the enum, the class, `get_tokenizer_type`, and the `TextEmbedder` constructor. Nothing existing changes behaviour.

Everything downstream already works unmodified: the ONNX export has two inputs (`input_ids`, `attention_mask`), and `batch_encode` pads ids with 0, which is Gemma's pad token.

## Verification

Built and tested with `docker/development-macos.Dockerfile` (bazel 6.2.0, aarch64, gcc-10).

**Builds clean.** `//:typesense-server` compiles with the existing `-Wall -Wextra -Werror=return-type`, no new warnings from these files.

**Unit tests pass.** `bazel test --test_arg=--gtest_filter=GemmaTokenizerTest.* //:typesense-test` — 5 tests, all passing. They cover raw ids (asserted against `SentencePieceProcessor` directly, so a fairseq offset fails the test), case preservation, BOS/EOS wrapping, the empty input, and truncation keeping EOS last. The fixture is a 1.1 KB SentencePiece model trained with Gemma's special-token layout, so no model download is needed.

I mutation-tested those cases to check they actually bite — a `+1` offset, lowercasing, a missing BOS, a missing EOS, an off-by-one truncation limit in either direction, and no truncation at all are each caught.

**End to end against the real binary.** Ran the built `typesense-server` with an EmbeddingGemma ONNX export in `<data-dir>/models/`, `model_type: "gemma"`. The collection creates with `num_dim` 768, documents index, and embeddings match `sentence-transformers` at **cosine 1.000000** on six documents spanning Latin, Chinese and German text. Cross-lingual search resolves correctly — "What is the capital of Japan?" ranks 东京是日本的首都。first at distance 0.411, "programozási nyelv" ranks the Python sentence first.

Separately, the tokenizer output was diffed against `transformers`' `AutoTokenizer` on 15 inputs (Latin, Chinese, Cyrillic, Arabic, Japanese, emoji, mixed case, whitespace, numerics) plus inputs of exactly 2044/2045/2046/2047/2048/2050 pieces to pin the truncation boundary. All ids and attention masks match exactly.

One packaging note for anyone exporting a model for this: `torch.onnx.export` emits IR version 10 and the pinned onnxruntime (`rel-1.14.1`) accepts at most IR 8, so `ir_version` has to be set back to 8. Opset 18 itself is fine.

## Two things I ran into

Neither is caused by this change, and I have left both alone.

- `src/text_embedder_tokenizer.cpp:86` reads `fairseq_tokens_to_ids_["<eos>"]`, but that map holds `</s>`, not `<eos>`. `operator[]` inserts the missing key with value 0, so a truncated XLM-RoBERTa input ends in token 0 (`<s>`) rather than 2 (`</s>`). This affects every `xlm_roberta` model currently shipped.
- `download_public_model` fetches only `model.onnx` and `model.onnx_data`, so an export whose external data is split across more files initialises locally but fails as a public model. Worth validating, or documenting alongside the undocumented `data_md5` config key.

The model itself is packaged in https://huggingface.co/typesense/models-moved/discussions/19, which is blocked on this PR.

https://claude.ai/code/session_01A9ydwqcWZWFG2bLUnDLt8E


honzasterba:
Two corrections to what this PR originally said.

**I can build locally after all.** The first version of this description claimed I could not build on Apple Silicon because the images in `docker/` are amd64. That was wrong — `docker/development-macos.Dockerfile` installs the arm64 bazel and runs natively. Everything is now built and tested with it, and the description above reflects real build and test output rather than standalone verification.

**Good catch on the truncation, thank you.** Fixed in 9d3c5d79 — `Encode` now bounds the copy up front:

```cpp
const size_t num_pieces = std::min(piece_ids.size(), max_length_ - 2);
input_ids.reserve(num_pieces + 2);
```

Output is unchanged. I diffed it byte-for-byte against the previous version on the 15-input corpus, and added inputs of exactly 2044, 2045, 2046, 2047, 2048 and 2050 pieces plus the empty string to pin the boundary — all still match `AutoTokenizer`. The unit tests added in c435c658 cover the boundary, and I checked that an off-by-one in either direction fails them.

honzasterba:
Full suite result, for completeness: **1590 passed, 21 failed** (`bazel test //:typesense-test`, aarch64, gcc-10). `GemmaTokenizerTest` passes inside that run.

I checked the 21 against base `245ebc7e` with this change absent rather than assuming they were pre-existing:

- **20 of them fail identically on base.** 14 × `Personalization*` (they fetch and unpack model archives; one fails with `attempting to parse an empty input`, i.e. an empty download), `RaftServerTest.ResolveNodesConfigWithIPv6` (`Unable to resolve host: ipv6.test-ipv6.com` — no IPv6 in the container network), plus `CollectionSortingTest.TestVectorQueryDistanceThresholdSorting`, `CollectionVectorTest.HybridSearchAuxScoreTest` and `PersonalizationAutoEmbeddingTest.TestAutoEmbeddingFields`. All environmental here.
- **`UnionTest.UnionHighlightingUAFRaceASAN` is flaky.** It happened to pass on base in the first comparison run, so I ran it repeatedly instead of drawing a conclusion from one sample: 3/5 pass on this branch, 4/8 pass on base. Nondeterministic either way, and unrelated to this change.

Nothing in the failing set touches the text embedder path.

PR #3039, opened 2026-09-06
Add a qwen tokenizer type for Qwen3-Embedding models

honzasterba:
Adds a `qwen` tokenizer type so the `Qwen3-Embedding` models can be used as built-in models.

## Why a new tokenizer is needed

Qwen3-Embedding uses byte level BPE. None of the existing tokenizers can express it:

| Path | Why it does not fit |
|---|---|
| `bert`, `distilbert` | need a WordPiece `vocab.txt` |
| `xlm_roberta`, `siglip` | need a SentencePiece model |
| `clip` | is byte level BPE, but with CLIP's own vocabulary, lowercasing and a 77 token limit |

## The change

`QwenTokenizer` reads the Hugging Face `tokenizer.json`. That file carries both the vocabulary and the merges, so it fits the single `vocab_file_name` that a model's `config.json` already allows — no new config key.

Encoding follows the pipeline declared in that file:

1. NFC normalise
2. split on the Qwen pre tokenization pattern, using ICU's regex engine
3. map each byte through the GPT-2 byte encoder
4. apply BPE merges in rank order
5. append `<|endoftext|>`, as the post processor template does

Four call sites, as with the other tokenizers: the enum, the class, `get_tokenizer_type`, and the `TextEmbedder` constructor. Nothing existing changes behaviour.

Two details worth pointing at in review:

**Merges are keyed on the two halves separated by a NUL**, not concatenated. Concatenation cannot distinguish `("ab","c")` from `("a","bc")`, and with 151k merges over a byte level alphabet those collisions do occur. The byte encoder maps every byte to a printable code point, so a NUL can never appear inside a token. `test/text_embedder_tokenizer_test.cpp` covers this directly. (The same pattern in `clip_tokenizer_cpp` concatenates; I have not touched it.)

**The tokenizer takes its own lock.** `batch_encode` is called outside the embedder's mutex, and the ICU matcher and the BPE cache are shared mutable state — the same reason `CLIPTokenizerWrapper` holds one. The BPE cache is also bounded, so a stream of unique tokens cannot grow it without limit.

## The 2048 token limit is measured, not inherited

Qwen3-Embedding accepts 32k tokens, but attention is quadratic and this runs on CPU. Measured on the exported model, batch of one:

| tokens | wall | peak RSS |
|---:|---:|---:|
| 128 | 0.20s | 1.7 GB |
| 512 | 1.24s | 1.8 GB |
| 1024 | 2.78s | 2.0 GB |
| 2048 | 7.06s | 2.6 GB |
| 4096 | 19.70s | 4.7 GB |
| 8192 | 65.68s | 10.4 GB |

A single 8192 token document would stall an indexing thread for over a minute and take 10 GB. 2048 keeps the worst case bounded and matches the cap the other recent tokenizers use. Happy to change it, or to make it configurable, if you would rather.

## Verification

Built and tested with `docker/development-macos.Dockerfile` (bazel 6.2.0, aarch64, gcc-10).

**Unit tests.** `bazel test --test_arg=--gtest_filter=QwenTokenizerTest.* //:typesense-test` — 10 tests, all passing. The fixture is a 3 KB `tokenizer.json` holding the 256 byte symbols and ten merges, chosen so the tests can distinguish rank ordered merging from leftmost merging, and a correct merge key from a concatenated one. No model download needed.

I mutation tested them: dropping the EOS, merging leftmost instead of lowest rank, concatenating the merge key, lowercasing, skipping the byte encoder, skipping NFC, an off by one truncation, and dropping the pre tokenization regex are each caught.

**Differential against `transformers`.** The `QwenTokenizer` declaration and definition, compiled verbatim out of this branch, were diffed against `AutoTokenizer` for `Qwen/Qwen3-Embedding-0.6B` on **1225 inputs** — Latin, Chinese, Cyrillic, Arabic and Japanese text, emoji, contractions, mixed case, whitespace runs, control characters, plus 1200 randomised fuzz cases. Every id matches, including the truncation path.

**ICU regex equivalence.** Since the pre tokenization pattern uses an inline case insensitive group and a negative lookahead, I checked ICU's engine against Python's `regex` module separately on **830 inputs** before relying on it: identical splits on all of them.

**End to end.** Feeding the ids from the compiled tokenizer through an ONNX export of Qwen3-Embedding-0.6B and applying `TextEmbedder::mean_pooling` reproduces `sentence-transformers` at **cosine 1.000000**, both for single inputs and for a right padded batch of six.

The model packaging is a separate PR against the models repo, which depends on this one.

https://claude.ai/code/session_01A9ydwqcWZWFG2bLUnDLt8E


honzasterba:
All three addressed in 7b0392d7, thank you.

**The malformed merge was worse than it looked.** The `continue` skipped the `rank++` as well, so it was not just dropping the bad entry — every merge after it loaded one rank too low, which changes how the whole vocabulary tokenizes. Both a merge of the wrong shape and a string with no separator now throw, and `rank` advances in step with the file either way:

```cpp
} else {
    // Skipping the entry would shift the rank of every merge after it,
    // which silently changes how the whole vocabulary tokenizes.
    throw std::runtime_error("Tokenizer file has a malformed merge at rank " +
                             std::to_string(rank) + ": " + vocab_path);
}
```

Two tests cover it (`RejectsMergeThatIsNotAPair`, `RejectsMergeOfTheWrongShape`), and I checked they fail against both the old `continue` and against a version that accepts the entry while keeping rank in step.

**The `const` return is gone** from both the declaration and the definition. You are right that it is misleading; I had been matching `XLMRobertaTokenizer::tokenize` and `token_to_id` just above, which do the same thing. I have left those alone rather than widen this PR, but happy to clean them up here or separately if you would like.

Rechecked after the changes: 12 unit tests pass under `bazel test`, and the differential run against `transformers` is still an exact match on all 1225 inputs.

PR #3049, opened 2026-09-15
perf: compute a lopsided `&&` from its selective side alone

alnr:
## What this does

`filter_by=A && B` costs what its *wider* side costs, regardless of what the conjunction matches. `filter_result_iterator_t::compute_iterators()` materialises both subtrees and intersects the two id arrays, so a filter that narrows to a hundred documents still pays for the leaf that matches two million.

The node's estimate is already correct — `min(left, right)` really is the size of the result — but a small result says nothing about the cost of *computing* it, and that cost is bounded by the larger side.

An `&&` yields at most as many ids as its narrower side. When the other side is at least `AND_PROBE_RATIO` (8) times wider, this materialises the narrow side only and asks the wide side about each id it yields, via the existing `is_valid(id)` — the same call `and_filter_iterators()` already uses to advance a lazy subtree. Same ids, same order, same `approx_filter_ids_length`. It is a different plan for the same query, not a change in semantics. No new flag.

## Scope

Pays off where the wide side is still lazy when its parent `&&` is initialised:

- any string leaf above `string_filter_ids_threshold`, in every configuration;
- an integer or float leaf when `enable_lazy_filter` is on.

A range-index, geo or `id:*` leaf has already materialised inside its own `init` before the `&&` node exists. Probing it is still correct — each probe is then a binary search — but it cannot give back a cost paid one level down.

Falls back to intersecting both sides when either subtree filters on a referenced collection (intersecting drops an id whose reference results don't intersect; probing has no such step), when the node is an object filter root, and when the wide side is an operator rather than a leaf.

Timeouts are unchanged: both children get a copy of the budget before anything is computed, the counted check inside `is_valid` cuts the probe loop short, and a forced check after the loop covers a loop too short to reach the clock. A `-1` from `is_valid` is *not* read as a timeout — in a lopsided `&&` it almost always means the wide side ran out of ids, which is an ordinary complete result — so `validity` distinguishes the two.

## Benchmarks

Apple M1 Max, 10 cores, 64 GB, macOS 26.6.2. Release builds (`bazel build //:typesense-server`, `-O2`). `origin/v31` vs this branch, same on-disk index, single request, no concurrency. `found` is identical in every row.

### `musicbrainz-1M-songs` — the corpus `benchmark/` already downloads

| scenario | found | before | after | speedup |
| --- | --- | --- | --- | --- |
| `filter_selective_and` (added here) | 431 | 31 ms | 1 ms | **31x** |
| `filter_complex` (existing) | 1,258 | 8 ms | 2 ms | **4x** |
| `filter_simple` (existing, single leaf — control) | 214,468 | 15 ms | 15 ms | 1x |
| `release_group_types:[Album,Single,Compilation]` alone | 960,372 | 61 ms | 61 ms | 1x |

`filter_complex` already exercised this path by accident: `&&` and `||` are left-associative, so it parses as `(genres:Rock && primary_artist_name:Queen) || primary_artist_name:Led Zeppelin`, and that inner conjunction is 960 against 214,468.

### Synthetic, 2M documents

Two string fields: one whose value matches 100 documents (the selective side), one whose value matches 1,900,000 (the broad side).

| `filter_by` | found | before | after | speedup |
| --- | --- | --- | --- | --- |
| selective `&&` broad | 95 | 92 ms | 1 ms | **92x** |
| broad `&&` selective | 95 | 92 ms | 1 ms | **92x** |
| selective leaf alone | 100 | 0 ms | 0 ms | 1x |
| broad leaf alone | 1,900,000 | 137 ms | 145 ms | 1x |
| ratio 2x between sides | 1 | 0 ms | 0 ms | 1x |
| ratio 8x between sides | 1 | 0 ms | 0 ms | 1x |
| ratio 1000x between sides | 5 | 5 ms | 0 ms | 5x |
| numeric wide side, default config | 95 | 37 ms | 38 ms | 1x |
| no `filter_by` | 2,000,000 | 0 ms | 0 ms | 1x |

The conjunction reaches its selective leaf from either operand order. The broad leaf's 137 → 145 ms is run-to-run noise measured minutes apart; a single leaf never reaches the operator branch that changed. The numeric wide side is unchanged, which is the scope limit above, not a broken probe.

## Tests

`compute_iterators()` deletes both subtrees before returning, so a private `computed_by_probe` with a `_get_computed_by_probe()` accessor is what lets a test assert which plan ran.

- `AndProbeSelectiveSide` — result equivalence for both operand orders, a wide side that runs out of ids first, a `!=` wide side, three leaves, an operator on either side, and a side matching nothing or everything.
- `AndProbeNumericSide` — numeric wide side, lazy and eager; the eager case documents the scope limit.
- `AndProbeRatioBoundary` — 39 vs 40 ids against a narrow side of 5.
- `AndProbeTimeout` — a probe loop cut short by the counted check, one too short to reach the clock and left to the forced check, and two cases without a budget that must *not* report a timeout, including the one where the wide side runs out of ids and returns the same `-1` a timeout does.
- `AndProbeKeepsReferences` — a joined collection still intersects and keeps its reference results.
- `CollectionFilteringTest.SelectiveAndDoesNotMaterializeWideSide` — end to end, sized around the thresholds a test build inverts.

Full suite run before and after: the same 21 tests fail on this machine either way (macOS cgroup parsing, ONNX model tests, NL-search model manager, metrics quantiles), so no regression. `FilterTest.ObjectFitlterIterator` and `FilterTest.FilterReferences` pass unchanged.


kishorenc:
My concern with the 1:8 ratio heuristic is that it also selects probing when both sides are large, e.g. 1M matches AND 9M matches means up to 1M lookups, which could be slower than a sequential intersection, especially if the larger side is already materialized.

There is an earlier PR by @tharropoulos in #2981 that gets this aspect right by requiring the smaller side’s estimated count to be below COMPUTE_FILTER_ITERATOR_THRESHOLD (25,000). Could we retain #3049’s implementation and safeguards, but adopt that cap alongside the 8× ratio, and skip probing when the larger side is already materialized Checking the actual smaller-side count after computing it would also protect against underestimated counts.

Please also benchmark 1M versus 9M against the existing intersection path, with both a lazy string broad side and an already-materialized numeric broad side (enable_lazy_filter=false). Clustered and scattered candidate IDs, with latency and peak-memory measurements, would help establish where probing pays off.

tharropoulos:
I can port the changes from here onto my PR

tharropoulos:
Benchmarked inside a container on Ubuntu 20.04, aarch64, using production `-O2` builds. Measurements include iterator initialization and computation, excluding indexing, HTTP, ranking, and serialization.

For 1M candidates versus 9M broad matches, every plan returned the same 900,000 IDs. `enable_lazy_filter=false` throughout.

| Broad side | Candidate IDs | Existing intersection (ms / MiB) | Ratio-only probing (ms / MiB) | Revised implementation (ms / MiB) |
|---|---|---:|---:|---:|
| Lazy string | Clustered | 187.45 / 215.5 | 81.81 / 85.5 | 179.55 / 215.8 |
| Lazy string | Scattered | 191.01 / 215.5 | 112.13 / 85.5 | 180.65 / 215.8 |
| Materialized numeric | Clustered | 59.31 / 139.0 | 74.05 / 139.2 | 59.23 / 139.5 |
| Materialized numeric | Scattered | 61.64 / 139.0 | 77.60 / 139.2 | 59.77 / 139.5 |

For smaller candidate sets against the same 9M-match lazy string side:

| Candidates | Candidate IDs | Existing intersection (ms / MiB) | Revised implementation (ms / MiB) | Revised plan |
|---:|---|---:|---:|---|
| 1,000 | Clustered | 169.655 / 211.0 | 0.155 / 70.6 | Probe |
| 1,000 | Scattered | 172.504 / 211.0 | 2.462 / 70.6 | Probe |
| 24,999 | Clustered | 179.102 / 211.1 | 2.146 / 70.6 | Probe |
| 24,999 | Scattered | 172.095 / 211.1 | 22.916 / 70.6 | Probe |
| 25,000 | Clustered | 170.655 / 211.2 | 179.364 / 211.3 | Intersect |
| 25,000 | Scattered | 169.547 / 211.2 | 165.843 / 211.3 | Intersect |

Probing the already-materialized numeric side was ~25–26% slower at 1M candidates, without meaningful peak-memory savings. The revised implementation avoids that regression.

Below the cap, lazy-string probing reduces latency and saves approximately 140 MiB of peak RSS. Probing is still faster at 1M for the string workload, so 25,000 can be tweaked, or even be passed as a server parameter instead.

alnr:
@tharropoulos Nice work, I've come to essentially the same results, including the regressions in some cases.

The right move is probably to have a probing ratio and a max cutoff (likely >1M), as well as don't attempt probing for certain wide side scenarios.

alnr:
> ⚠️ **AI-generated analysis.** Written by Claude (Anthropic) running in Claude Code. The benchmarks below were run locally against this branch; corpus generator, harness scripts and raw results can be shared on request.

Following up on the 1:8 ratio concern with measurements. Short version: the deciding factor is **how many times bigger the broad side is**, the break-even sits near **40×**, and an absolute cap on the selective side cannot separate the profitable cases from the unprofitable ones at any value.

## Method

**Hardware/build.** Apple M1 Max (10 cores, 64 GB), macOS 26.6.2. Release builds (`bazel build //:typesense-server`, `-O2`). Single node, one request at a time, no concurrency. Timings are the server's own `search_time_ms`, median of 4 timed runs after a full warm-up pass over every scenario. Both plans query the same on-disk index.

**Two binaries**, so that every scenario is timed under *both* plans:

1. `origin/v31` — always intersects.
2. This branch built with `AND_PROBE_RATIO = 1` and no cap — probes whenever the sides differ at all.

Measuring both plans for every scenario means any candidate gate rule can then be evaluated arithmetically over the results, without rebuilding per parameter value.

**Corpus.** 10,000,000 documents, four string fields:

```json
{"id": "<i>", "n": "<class>", "w9m": "yes|no", "w1m": "yes|no", "w100k": "yes|no"}
```

* `n` is one of `n100`, `n1k`, `n10k`, `n25k`, `n100k`, `n500k`, `n1m`, `none` — the selective sides, with exactly 100 / 1,000 / 10,000 / 25,000 / 100,000 / 500,000 / 1,000,000 matching documents. Class membership is assigned from `(i * 2654435761) % 10_000_000`, which is a permutation of the id space, so each class has exactly its nominal count **and is spread uniformly over the ids** rather than sitting in one contiguous run. (An earlier version of this corpus used contiguous runs, which confounds probe cost with how far into the broad side's posting list the probes reach.)
* `w9m` = `yes` for `i % 10 != 0` → 9,000,000 documents; `w1m` = `yes` for `i % 10 == 0` → 1,000,000; `w100k` = `yes` for `i % 100 == 0` → 100,000. Also spread rather than prefixes, so the probed side never runs out of ids early.

**Queries.** All 7 × 3 combinations, plus each leaf alone to record exact match counts:

```
GET /collections/grid/documents/search
    ?q=*&per_page=1&filter_by=n:=<class> && <wide>:=yes
```

`q=*` so the filter dominates the search. All sides are string equality leaves; numeric and already-materialised sides are out of scope for this grid.

**Result equality.** All 21 scenarios returned identical `found` under both plans.

**One labelling note.** The engine computes the smaller side and probes the larger one, so "selective side" below is `min(side A, side B)` regardless of which field it came from. Two rows land on the same 100,000-vs-1,000,000 shape from different fields; they agree to within 0.2 ms, which is a useful replication check.

## The 21 scenarios

| selective side | broad side | broad side is | intersect | probe | result |
| ---: | ---: | ---: | ---: | ---: | --- |
| 100,000 | 100,000 | same size | 5 ms | 76 ms | **15× slower** |
| 1,000,000 | 1,000,000 | same size | 51 ms | 750 ms | **15× slower** |
| 500,000 | 1,000,000 | 2× | 38 ms | 387 ms | **10× slower** |
| 25,000 | 100,000 | 4× | 4 ms | 20 ms | **5× slower** |
| 100,000 | 500,000 | 5× | 14 ms | 78 ms | **5× slower** |
| 1,000,000 | 9,000,000 | 9× | 288 ms | 836 ms | **2.9× slower** |
| 10,000 | 100,000 | 10× | 5 ms | 16 ms | **3.2× slower** |
| 100,000 | 1,000,000 | 10× | 26 ms | 82 ms | **3.2× slower** |
| 100,000 | 1,000,000 | 10× | 27 ms | 81 ms | **3.0× slower** |
| 500,000 | 9,000,000 | 18× | 252 ms | 442 ms | **1.8× slower** |
| 25,000 | 1,000,000 | 40× | 24 ms | 26 ms | about the same |
| 100,000 | 9,000,000 | 90× | 225 ms | 126 ms | 1.8× faster |
| 1,000 | 100,000 | 100× | 4 ms | 2 ms | 2× faster |
| 10,000 | 1,000,000 | 100× | 46 ms | 27 ms | 1.7× faster |
| 25,000 | 9,000,000 | 360× | 220 ms | 49 ms | 4.5× faster |
| 10,000 | 9,000,000 | 900× | 437 ms | 45 ms | 9.7× faster |
| 100 | 100,000 | 1,000× | 4 ms | 0 ms | 8× faster |
| 1,000 | 1,000,000 | 1,000× | 47 ms | 4 ms | 12× faster |
| 1,000 | 9,000,000 | 9,000× | 435 ms | 5 ms | 87× faster |
| 100 | 1,000,000 | 10,000× | 45 ms | 0 ms | 90× faster |
| 100 | 9,000,000 | 90,000× | 436 ms | 0 ms | 872× faster |

`0 ms` rows are genuine — the API reports whole milliseconds and probing 100 ids takes less than one.

Running all 21 searches back to back: **always intersect 2634 ms**, **always probe 3052 ms**, **picking the better plan each time 992 ms**. The last figure is the yardstick for the tables below.

The two summary columns used throughout: *all queries* is the sum over the 21 searches, and *worst query* is, for the single worst search under that rule, how much slower it was than it needed to be — which covers both "took the slower plan" and "refused a plan that would have been faster".

## 1. Ratio only

| require broad ≥ | all queries | vs best possible | worst query | searches it gets wrong |
| --- | ---: | ---: | ---: | ---: |
| 8× (current) | 1853 ms | 1.87× | 3.2× | 5 |
| 12× | 1184 ms | 1.19× | 1.8× | 1 |
| 16× | 1184 ms | 1.19× | 1.8× | 1 |
| 24× | 994 ms | 1.00× | 1.1× | 0 |
| 32× | 994 ms | 1.00× | 1.1× | 0 |
| 40× | 994 ms | 1.00× | 1.1× | 0 |
| 48× | 992 ms | 1.00× | 1.0× | 0 |
| 64× | 992 ms | 1.00× | 1.0× | 0 |
| 128× | 1113 ms | 1.12× | 2.0× | 3 |

**What "gets wrong" counts.** Both plans were timed for every search, so one of them is objectively the better choice for that search; the column counts the searches where the rule picked the other one by more than 20%. The margin keeps ties and measurement noise from being counted as mistakes. Results are identical either way — the cost is time, never correctness. The mistakes come in two opposite directions:

* **Too permissive** — probed where intersecting was faster. All five mistakes at 8× are this kind: 10k vs 100k (3.2× off), 100k vs 1M (3.2×), 100k vs 1M (3.0×), 500k vs 9M (1.8×), 1M vs 9M (2.9×).
* **Too strict** — refused to probe where probing was faster. All three mistakes at 128× are this kind: 1k vs 100k (2.0× off), 10k vs 1M (1.7×), 100k vs 9M (1.8×).

Between 24× and 64× the count is zero in both directions.

## 2. Cap only (keeping the current 8× requirement)

| require selective < | all queries | vs best possible | worst query |
| --- | ---: | ---: | ---: |
| 1,000 | 2150 ms | 2.17× | 87× |
| 10,000 | 1674 ms | 1.69× | 9.7× |
| 25,000 | 1274 ms | 1.28× | 4.5× |
| 100,000 | 1104 ms | 1.11× | 3.2× |
| 500,000 | 1116 ms | 1.12× | 3.2× |
| no cap | 1853 ms | 1.87× | 3.2× |

**A cap cannot fix this at any value**, because the losing searches have *small* selective sides: 10,000 against 100,000, and 100,000 against 1,000,000, both 10× and both ~3× slower. No cap that admits the wins can exclude those. Meanwhile a cap low enough to bite throws away the wins it was never meant to touch — at 25,000 it refuses 25,000 vs 9,000,000 (4.5× faster) and 100,000 vs 9,000,000 (1.8× faster). Best case for a cap is 1.11× of optimal with a 3.2× worst case; it never reaches optimal.

In this table `worst query 87×` means a search that could have run in 5 ms was left at 435 ms — a refused win, not a slowdown.

## 3. Ratio and cap together

| | selective < 25,000 | selective < 100,000 | selective < 500,000 | no cap |
| --- | --- | --- | --- | --- |
| broad ≥ 8× | 1274 ms / worst 4.5× | 1104 ms / worst 3.2× | 1116 ms / worst 3.2× | 1853 ms / worst 3.2× |
| broad ≥ 16× | 1263 ms / worst 4.5× | 1093 ms / worst 1.8× | 994 ms / worst 1.1× | 1184 ms / worst 1.8× |
| broad ≥ 24× | 1263 ms / worst 4.5× | 1093 ms / worst 1.8× | **994 ms / worst 1.1×** | **994 ms / worst 1.1×** |
| broad ≥ 32× | 1263 ms / worst 4.5× | 1093 ms / worst 1.8× | **994 ms / worst 1.1×** | **994 ms / worst 1.1×** |
| broad ≥ 48× | 1263 ms / worst 4.5× | 1092 ms / worst 1.8× | 992 ms / worst 1.0× | **992 ms / worst 1.0×** |

Once the ratio requirement reaches 24×, the cap contributes nothing — the last two columns are identical. A 25,000 cap makes the result *worse* at every ratio (1263 ms against 994 ms), for the reason above.

## What the numbers support

* Profitability tracks the size ratio and is independent of absolute size: 100,000 vs 1,000,000 loses by 3× just as 1,000,000 vs 9,000,000 loses by 2.9×, and both are ~10×.
* Break-even is bracketed between 18× (1.8× slower) and 40× (a tie), with 90× already 1.8× faster. Anything from ~24× to ~48× scores identically on this grid, so **32×** sits mid-band with margin either side.
* A cap adds nothing once the ratio requirement is right, and at 25,000 it costs more than it saves.
* Why 8× looked harmless until now: the synthetic table in this PR's description measured its "ratio 8x" row as 100 documents against 800, which times at 0 ms → 0 ms. The relative penalty at 8× was always around 2.5–3×; it only becomes visible in wall-clock terms once the selective side is large enough for that multiple to matter.

## Caveats

* Summing 21 searches weights every shape equally, which no real workload does — the *worst query* column is the more useful one, and both are given.
* Whole-millisecond resolution flatters the extreme wins (872× is really "436 ms versus under a millisecond").
* Single request, no concurrency, single node, warm cache.
* String equality leaves only.
* The break-even is bracketed, not pinpointed: there are no measured shapes between 18× and 40×.

---

*Edited: the "searches it gets wrong" figure for the 8× row was 4 and is corrected to 5 — the missed case is 500k vs 9M (18× lopsided), 1.8× off. That count had been transcribed by hand; every other figure in these tables came straight from the measurement scripts. The explanation of that column has also been expanded.*

alnr:
> ⚠️ AI-generated summary of the changes in `8e9f96e9`, reviewed before posting.

Pushed `8e9f96e9`, following up on the ratio concern.

**What changed**

- `AND_PROBE_RATIO` 8 → **32**. Profitability tracks the ratio between the sides, not their absolute sizes — probing loses ~3× at 10× whether the narrow side holds 10k or 100k ids, breaks even near 32×, and wins by 9.7× at 900×.
- **Two wide sides now keep the intersecting plan unconditionally:** one that has already materialised its ids (nothing left to skip — intersecting is a linear merge, probing a binary search per id), and a lazy numeric range (one iterator per matching value, so a probe costs a pass over the range — 134 s against 7.6 s at 10M documents, and a truncated result once the budget ran out).
- The gate decides on `approx_filter_ids_length`; it is now **re-checked against the narrow side's exact count** once that side is computed, before any seek into the wide one.
- Benchmark scenarios added for the band the ratio decides (17× and 32×) and for a numeric wide side, materialised and lazy.

**On #2981**

Nothing from it landed. I ported its `init()` gate (deciding eager materialisation on the *larger* operand, with the zero-match short-circuit), measured it, and reverted: across four lazy-numeric scenarios it went 45.3 s → 72.1 s including one truncated result (10 rows where 102 were expected), while string shapes were unchanged (1047 ms vs 1070 ms). Staying lazy only pays when the larger side is cheap to step through, which a numeric range isn't. A variant restricting it to lazy string leaves still regressed two of the four and broke two tests.

Our reference handling remains stricter than #2981's — the probed side must be a leaf, and any referenced filter anywhere in the subtree refuses the plan.

**Against your four points**

| point | status |
| --- | --- |
| 1:8 admits both-sides-large | **Addressed** — confirmed (1M && 9M was 2.3× slower) and fixed by the ratio, not a cap |
| Adopt #2981's 25,000 cap | **Not adopted** — a cap can't separate the cases at any value: the losing shapes have *small* narrow sides too (10k vs 100k, 100k vs 1M, both 10×, both ~3× slower), and at 25,000 it refuses real wins (25k vs 9M is 4.5× faster). Sweep in the previous comment |
| Skip probing an already-materialised larger side | **Addressed** — guard added; note #2981 doesn't do this either |
| Check the actual smaller-side count after computing | **Addressed** |
| Benchmarks: 1M vs 9M, lazy string and eager numeric, clustered and scattered | **Addressed** — in the previous comment |
| …with peak-memory measurements | **Not addressed** — only RSS sampling that isn't comparable across runs; latency only |

**Caveat on the new benchmark scenarios:** on my hardware the 1M-song corpus can't resolve them — the unfiltered controls (`just_q`, `q_star`) swing 30–50% between runs, so anything below that is noise. They should be meaningful on quieter CI. The 10M-document measurements are the evidence for the threshold itself.

alnr:
I don't think the failing API tests are related to this PR but general flakes.

PR #3051, opened 2026-09-15
test: cover infix matches with a curated hit and a filter_by that matches nothing

alnr:
Additional regression tests for https://github.com/typesense/typesense/commit/d4dbdae2a108b81dc7bfae6a6fd02b5ab139d2ea


test/collection_specific_test.cpp
```diff
@@ -73,6 +73,41 @@ class CollectionSpecificTest : public ::testing::Test {
...
+                    {"name": "category", "type": "string", "facet": true}
+                ]
+            })"_json;
+        auto collection_create_op = collectionManager.create_collection(schema_json);
+        ASSERT_TRUE(collection_create_op.ok()) << collection_create_op.error();
+
+        auto collection = collection_create_op.get();
+        for(size_t i = 0; i < 12; i++) {
+            // "stock" matches these titles only as an infix, so only the infix search can find them.
+            nlohmann::json document = {
```
tharropoulos, line 91: could we assign an explicit id here with `{"id", std::to_string(i)}`? the tests pin 5 and hide 7, so making those ids explicit would keep that relationship clear without relying on automatic id assignment.

test/collection_specific_test.cpp
```diff
@@ -2702,6 +2737,57 @@ TEST_F(CollectionSpecificTest, GroupedHitsExcludePhraseMatchesWhenFilterMatchesN
...
+                               {"group_by", "category"}});
+
+    std::vector<std::string> ids;
+    for(const auto& group: result["grouped_hits"]) {
+        for(const auto& hit: group["hits"]) {
+            ids.push_back(hit["document"]["id"].get<std::string>());
+        }
+    }
+
+    ASSERT_EQ(std::vector<std::string>{"5"}, ids);
```
tharropoulos: Could we also assert that grouped_hits contains one group, that its group_key is ["garden"], and that found/found_docs are 1? Checking just the flattened ids would still pass with incorrect totals or more empty groups.

test/collection_specific_test.cpp
alnr: done

test/collection_specific_test.cpp
alnr: done, but slightly different assertion. Curated hits don't count against `found_docs`.

kishorenc (approved):

PR #3052, opened 2026-09-15
fix: report search_cutoff and 408 correctly for union searches

alnr:
- A union took `search_cutoff` from its last sub-search only, so an earlier sub-search running out of budget went unreported.
- A union cut off with nothing to return answered 200 with an empty body instead of 408: `CollectionManager::do_union` returned `result_op` rather than the 408, and the handler's union 408 branch never finalized its response.


test/core_api_utils_test.cpp
```diff
@@ -1602,6 +1602,57 @@ TEST_F(CoreAPIUtilsTest, Union) {
...
+
+    for(size_t i = 0; i < 200; i++) {
+        nlohmann::json doc;
+        doc["title"] = "widget " + std::to_string(i);
+        doc["category"] = "kitchen";
+        ASSERT_TRUE(coll1->add(doc.dump(), CREATE).ok());
+    }
+
+    std::shared_ptr<http_req> req = std::make_shared<http_req>();
+    std::shared_ptr<http_res> res = std::make_shared<http_res>(nullptr);
```
tharropoulos: this uses http_res(nullptr), so stream_response() returns immediately. the test would still pas if we removed the new streaming call. could we cover the actual response delivery too?

test/union_test.cpp
```diff
@@ -3208,3 +3234,42 @@ TEST_F(UnionTest, SearchCutoffKeepsInfixMatchesFiltered) {
...
+    // Begun with the budget already spent, which is the state every sub-search
+    // of a union is in once an earlier one has used it up.
+    req_params = {{"per_page", "20"}, {"search_cutoff_ms", "1"}};
+    auto search_op = collectionManager.do_union(req_params, embedded_params, searches, json_res,
+                                                now_ts - std::chrono::microseconds(std::chrono::seconds(1)).count());
+
+    // The first sub-search is cut off before it finds anything and the second
+    // has nothing its filter admits, so the union has nothing to return. It
+    // answers 408 only if the flag it reads says a sub-search was cut off.
+    ASSERT_FALSE(search_op.ok());
```
tharropoulos, line 3253: could we also keep a pinned hit in the first search and assert `search_cutoff: true` when the last     search matches nothing?

test/union_test.cpp
```diff
@@ -434,6 +434,32 @@ class UnionTest : public ::testing::Test {
         }
     }
 
+    // Two collections for one union: the first holds documents the search has
+    // to work through, the second holds none that filter_by admits, so its
+    // search returns before it ever reads the clock.
+    void setupCutoffCollections() {
```
tharropoulos: this only used once, in one test, so this can be inlined in the test itself

src/collection.cpp
```diff
@@ -4184,6 +4184,11 @@ Option<bool> Collection::do_union(const std::vector<uint32_t>& collection_ids,
     auto found_docs = 0;
     std::unordered_map<uint32_t, std::unique_ptr<id_list_t>> union_result_seq_ids_by_collection;
 
+    // Every sub-search clears search_cutoff before it runs, so the flag left
+    // behind belongs to the last one. The union spends one budget across all of
+    // them and reports one flag, which has to mean any of them ran out of it.
```
tharropoulos: Super pedantic and could be ignored, but this could be rewritten as:

```
// each sub-search resets search_cutoff. preserve if any search was cut off.
```

alnr: done

test/union_test.cpp
alnr: done

test/union_test.cpp
alnr: Added, and restructured so the tests no longer depend on the last sub-search happening not to read the clock. Only the first sub-search has its budget spent (`search_cutoff_ms: 0` in its own body); the second runs with the default budget and returns real hits. The union has to report `search_cutoff: true` with `found` exactly 1 pinned + 20 real hits.

test/core_api_utils_test.cpp
alnr: Restructured, and added an API-level test for this. Fixed some related bugs at the same time.

src/core_api.cpp
```diff
@@ -1148,9 +1148,11 @@ bool post_multi_search(const std::shared_ptr<http_req>& req, const std::shared_p
     if (is_union) {
         Option<bool> union_op = CollectionManager::do_union(req->params, req->embedded_params_vec, searches,
                                                             response, req->conn_ts, union_remove_duplicates);
-        if(!union_op.ok() && union_op.code() == 408) {
+        if(!union_op.ok()) {
```
tharropoulos: could we keep the original 408 condition here? `do_union` only returns a failed option for 408 right now, so the broader check and nested `if` don’t change behavior.

api_tests/tests/union_search_cutoff.test.ts
```diff
@@ -0,0 +1,117 @@
...
+        query_by: "title",
+        filter_by: "category:=kitchen",
+        ...secondSearchParams,
+      },
+    ],
+  });
+}
+
+describe(Phases.SINGLE_FRESH, () => {
+  it("create the union cutoff collections", async () => {
```
tharropoulos: this could be changed for a `beforeAll` hook instead of a test

alnr:
@tharropoulos can I ask for a quick re-review?

src/core_api.cpp
alnr: done

api_tests/tests/union_search_cutoff.test.ts
alnr: done

tharropoulos (approved):
CC: @kishorenc :shipit:

Issue #3029, opened 2026-08-20
Synonym matching fires on unrelated queries when two synonym terms share a prefix longer than MAX_PREFIX_LEN (8 bytes)


thebossza101:
### Description

When a synonym set contains two terms where one is a **prefix of the other and the shared prefix is longer than `MAX_PREFIX_LEN` (8 bytes)**, synonym matching starts firing for unrelated queries. Any query token that differs from the synonym term only *after* the 8th byte is treated as an exact match, so the synonym expands and floods the result set.

This is **not** typo tolerance: it reproduces with `num_typos=0`, `synonym_num_typos=0` and `synonym_prefix=false`. None of those parameters have any effect on it.

The impact is much larger for languages whose characters take multiple bytes in UTF-8. A Thai character is 3 bytes, so a 3-character Thai word already exceeds the 8-byte limit and virtually every Thai synonym is affected. We hit this in production: the synonym set `["กระดาษ", "กระดาษถ่ายเอกสาร", "กระดาษA4"]` (paper / copier paper / paper A4) made the unrelated queries `กระดาน` (whiteboard), `กระบอก` (cylinder), `กระดูก` (bone), `กระติก` (flask) and `กรรไกร` (scissors) all return the ~640 paper products instead of their own results — a query for "whiteboard" returned nothing but paper.

The repro below uses ASCII only, so no multi-byte handling is needed to see it.

### Steps to reproduce

Typesense **v30.2** (also present in v30 and v31 source).

```bash
export TS="http://localhost:8108"
export KEY="xyz"

# 1. collection with a plain string field (default locale)
curl -s "$TS/collections" -X POST \
  -H "X-TYPESENSE-API-KEY: $KEY" -H 'Content-Type: application/json' -d '{
    "name": "prefix_bug",
    "fields": [{"name": "name", "type": "string"}]
  }'

# 2. documents
curl -s "$TS/collections/prefix_bug/documents/import?action=create" -X POST \
  -H "X-TYPESENSE-API-KEY: $KEY" --data-binary '{"name":"bookkeeping"}
{"name":"bookkeepinh"}
{"name":"bookkeepingx"}
{"name":"book"}
{"name":"bool"}'

# 3. synonym set: the two terms share an 11-byte prefix ("bookkeeping")
curl -s "$TS/synonym_sets/pfx" -X PUT \
  -H "X-TYPESENSE-API-KEY: $KEY" -H 'Content-Type: application/json' -d '{
[...]

PR #3054, opened 2026-09-18
Fix synonym matching for long ART prefixes

krunal1313:
### Problem

  Fuzzy synonym matching could return incorrect matches for long tokens that shared a prefix longer than ART’s `MAX_PREFIX_LEN`.

  ART stores only the first `MAX_PREFIX_LEN` bytes of compressed intermediate-node prefixes. During fuzzy traversal, bytes beyond that stored prefix were treated as matching
  the query, allowing differences in the omitted portion to go undetected.

  This was particularly visible with:

  - Long synonym terms
  - synonym_num_typos: 0
  - Multibyte languages such as Thai
  - Multiple synonym terms sharing a long prefix

  ### Fix

  The fix remains outside the generic ART implementation and is limited to synonym matching:

  - ART continues to generate ordered candidate leaves.
  - Long candidates are revalidated using their complete leaf keys.
  - Validation uses the same byte-level Optimal String Alignment distance semantics, including adjacent transpositions.
  - synonym_prefix matches are validated against candidate prefixes.
  - Exact matches retain the existing fast path.
  - Short-only synonym nodes retain the existing ART behavior.
  - The existing maximum of 10 returned synonym children remains enforced.

  If invalid long-prefix candidates consume the initial ART result window, the candidate window is progressively expanded until:

  - 10 valid matches are found,
  - ART candidates are exhausted, or
  - all children have been considered.

  Per-node metadata avoids scanning every synonym child on each query.
## PR Checklist


kishorenc (approved):

krunal1313:
Fixes #3029

Issue #3027, opened 2026-08-19
Failed async-reference alias swap rollback is not persisted across restart


cyppe:
## Summary

When an alias used by an `async_reference` is swapped to an incompatible collection, Typesense returns an error and appears to roll the alias back in memory. However, the rejected alias target is persisted. After restarting Typesense, the alias points to the rejected target and JOINs using the reference break.

This was reproduced twice, including once in a fresh isolated container.

## Environment

- Typesense: `31.0.rc14`
- Docker image: `typesense/typesense:31.0.rc14`
- Image digest: `sha256:6f63c3de844ce3c399dee23e04530f38e80a7462de621f1c8444245325320baa`
- Single-node Docker deployment

## Reproduction

1. Create `parent_v1` with a `code` string field.
2. Create alias `parent -> parent_v1`.
3. Create `children` with:

```json
{
  "name": "parent_code",
  "type": "string",
  "reference": "parent.code",
  "async_reference": true
}
```

4. Index `parent_v1` document `{"id":"p1","code":"ok"}` and child document `{"id":"c1","parent_code":"ok"}`.
5. Create `parent_bad` with only an `other` string field, so it does not contain the referenced `code` field.
6. Attempt to swap the alias:

```http
PUT /aliases/parent

{"collection_name":"parent_bad"}
```

Typesense returns HTTP 500:

```json
{"message":"Referenced field `code` not found in the collection `parent_bad`."}
```

7. Before restart, verify that rollback appears successful:

- `GET /aliases/parent` returns `parent_v1`.
- The JOIN still returns the child and referenced parent.
- The child helper field still resolves to sequence ID `0`.

8. Restart Typesense and check again.

## Actual behavior after restart

- `GET /aliases/parent` now returns `parent_bad`.
- The child's `parent_code_sequence_id` becomes the unresolved sentinel `4294967295`.
- The JOIN fails with:

```text
Referenced collection `parent` not found.
```

Relevant startup log:

```text
Field `code` not found in the collection `parent_bad` which is referenced in `children.parent_code`.
```

## Code pointer / possible fix direction

The risky ordering appears to be in `CollectionManager::upsert_symlink()`:

- The candidate alias target is [written to the store and published in memory](https://github.com/typesense/typesense/blob/ee7784f33195851f98bbc38b43696999543d7492/src/collection_manager.cpp#L1341-L1359).
[...]

PR #3055, opened 2026-09-18
Fix alias replay ordering

krunal1313:
## Change Summary

An alias used by an async_reference could be swapped to an incompatible collection, return an error, appear rolled back, then point to the rejected target after restart.
  This broke async-reference helper resolution and JOIN queries.

  Root cause: alias upsert requests had no BatchedIndexer identity. During replay, they could run before collections that reference the alias had been created. With no
  reference metadata available, validation was skipped and the invalid target was persisted.

  Fix:

  - Queue alias upserts by alias name, consistently during enqueue, replay, and completion.
  - Add a scheduler dependency from alias to requested target collection.
  - Let existing reverse-reference dependency tracking make alias updates wait for alias-referencing collections.
  - Refresh alias dependency state from the committed alias target after request completion, so failed swaps retain the old target relationship.

  Coverage:

  - Scheduler regression verifies alias route identity, target dependency, restored-request identity, and ordering behind referencing collections.
  - C++ restart regression verifies alias persistence, async helper sequence ID, and JOIN behavior after a rejected swap.
  - HTTP API regression recreates the reporter’s single-node flow and verifies the old alias target and JOIN after restart.

## PR Checklist


krunal1313:
Fixes #3027

PR #3057, opened 2026-09-19
perf: pool curl handles with shared dns and tls sessions on curl 8.22

tharropoulos:
- bump bundled curl 7.87 → 8.22, dropping the ~1s per-connect stall introduced by the 7.86/7.87 connection-filter refactor [1]
- lease easy handles from a lifo pool in `init_curl`. The sync path reuses live connections across request threads. Each search runs on a fresh thread, so `thread_local` caching never hit. handles return through `curl_easy_reset`, which keeps live connections, the dns cache and tls sessions on the handle [2]
- share dns and tls sessions process-wide via `curl_share` lock callbacks [3]. connections stay per handle since libcurl cannot share a connection cache across concurrent threads [4][5]
- negotiate http/2 with http/1.1 fallback (`CURL_HTTP_VERSION_2TLS`) on https and keep h2c prior knowledge on plain http, since curl 8.10 changed prior knowledge to offer only h2 in alpn, which breaks http/1.1-only model endpoints [6][7]
- compare url schemes case-insensitively. `HTTPS://` urls no longer skip `configure_ssl` ca and verify-mode setup (pedantic, but saves issues in the future)
- **IMPORTANT** unblocks jev. outbound llm reconnects dominated every nl search, dropping from 1851ms → ~355ms per call and 12 requests going from 12 syns to 0

Tested against openai text-embedding-3-small and gpt-4o-mini (5 iterations per case, packet capture per phase. `old` = v31 with curl 7.87 and no pooling, `new` = this branch).

```
variant   scenario     n ok    wall ms min/med/max tls-hs ms httpver  syn hello  dns
------------------------------------------------------------------------------------
direct    embed_cold   5  5            208/241/265        19       2    5     5   10
direct    embed_warm   5  5            176/209/332         0       2    1     1    2
direct    chat_cold    5  5           593/638/1675        21       2    5     5   10
direct    sse_cold     5  5            543/583/760        19       2    5     5   10

old       embed        5  5            458/471/721         -       -    5     5   15
old       conv_sse     5  5         1117/1208/1284         -       -    5     5   15
old       nl_search    5  5            886/944/992         -       -    5     5   10

new       embed        5  5            174/179/245         -       -    0     0    5
new       conv_sse     5  5           849/966/1660         -       -    5     5   15
new       nl_search    5  5            623/778/982         -       -    0     0    0
```

Legend: `syn` counts new tcp connections to :443, `hello` counts tls clienthellos and `dns` counts resolver queries inside each scenario's window. `tls-hs` is curl's `time_appconnect - time_connect`.

Once the pool is warm, embedding and nl search stop opening connections altogether and only model compute remains afterwards. A pooled call is as fast as a keep-alive curl process hitting the same endpoint. Conversational sse improves on the new curl but still does a fresh handshake on every request because streaming handles are neither pooled nor on the share. Attaching the share to them could be added as the next PR.

[1] https://curl.se/ch/7.87.0.html
[2] https://curl.se/libcurl/c/curl_easy_reset.html
[3] https://curl.se/libcurl/c/libcurl-share.html
[4] https://curl.se/libcurl/c/curl_share_setopt.html
[5] https://curl.se/libcurl/c/threadsafe.html
[6] https://curl.se/libcurl/c/CURLOPT_HTTP_VERSION.html
[7] https://curl.se/ch/8.10.0.html

## PR Checklist


tharropoulos:
Reference to [d410b73](https://github.com/typesense/typesense/pull/3057/commits/d410b73fdfd349e27ed00551a55ffb0d1b00616b), the build changes make curl’s previous behavior explicit by disabling Brotli and Zstandard in the Bazel curl build. In curl 7.87.0, both [`CURL_BROTLI` and `CURL_ZSTD` defaulted to `OFF`](https://github.com/curl/curl/blob/curl-7_87_0/CMakeLists.txt#L729-L741), and we weren’t enabling curl compressed-response decoding in the HTTP client.

kishorenc (approved):

Issue #2952, opened 2026-06-11
A non-selective `range_index` numeric filter is fully materialized on every search and is not bounded by `search_cutoff_ms`, so concurrent searches OOM-kill the node


alangmartini:
## Bug Description

When a search's `filter_by` contains a numeric `range_index` clause that is **non-selective** (a `>`/`>=`/`<` comparison that matches almost every document, e.g. `expires_at:>NOW` on data where nothing has expired yet), Typesense materializes the full matching id array on **every request**, even when the same `&&` (AND) expression contains a far more selective sibling clause that could drive the iteration instead.

Three things compound:

1. **The non-selective clause is always fully materialized.** Per-request memory is proportional to the cardinality of that clause (millions of ids), and is essentially independent of how many documents the overall query returns (a query returning 0–few hits costs the same).
2. **`search_cutoff_ms` does not bound the filter-build phase.** The cutoff is only checked while iterating results, not while the filter id set is being materialized, so a heavy filter build runs to completion regardless of the configured cutoff.
3. **The OOM resource guard covers writes only.** `memory-used-max-percentage` rejects writes when memory is high, but the search path is not gated, so concurrent heavy searches can drive the process past physical RAM.

Together, a burst of concurrent searches that each carry a per-request value in the non-selective clause (so the result cache never hits) accumulates `in_flight × per_request_alloc` and **OOM-kills the node**. A single such query is cheap enough to look harmless; the failure only appears under concurrency.

## Reproduction Steps

Self-contained: starts a memory-capped Typesense in Docker, loads 2M synthetic docs (all `curl` for Typesense calls; `python3` only to generate the JSONL and drive concurrency), then ramps concurrent searches until the container is OOM-killed. Deterministic — OOM-kills at 40 concurrent workers on a 1.5 GB cap.

```bash
bash repro.sh
```

<details>
<summary>repro.sh (full self-contained reproducer)</summary>

```bash
#!/usr/bin/env bash
# A non-selective range_index numeric filter is fully materialized on every search
# (ignoring a more selective AND-sibling), and search_cutoff_ms does not bound the
# filter-materialization phase. Under a burst of concurrent searches the per-request
# allocation accumulates and OOM-kills the node.
#
# Requires: docker, curl, python3 (only to generate synthetic JSONL + drive concurrency).
# Usage: bash repro.sh
set -u
NAME=ts-repro
KEY=localkey
PORT=8108
URL="http://localhost:${PORT}"
[...]

PR #3063, opened 2026-09-22
Reduce memory usage of nested numeric range filters

krunal1313:
## Summary

  Fix excessive intermediate memory usage in nested numeric range_index filters, including when enable_lazy_filter is omitted or false.

  - Reuse existing numeric posting-list iterators instead of eagerly expanding broad ranges.
  - Defer nested filter materialization until the final intersection.
  - Preserve cancellation, negation, JOIN/object semantics, sorting, facets, grouping, and curated results.
  - Remove unnecessary per-advance filter and schema copies.

  No public API, schema, or lazy-filter default changes.

  ## Validation

  - 412 regression tests passed before the final copy optimization; 108 focused tests passed afterward.
  - Debug server build and independent reviews passed.
  - Unprofiled baseline/candidate comparison: 4,800/4,800 requests passed, covering omitted/false/true lazy-filter modes.
  - On the two-million-document, 40-worker fixture, peak cgroup memory decreased 67–72% and p95 latency decreased 19–24%.

  Neither build OOMed in this fixture; results demonstrate reduced memory pressure, not universal OOM prevention.

  Addresses #2952.
## PR Checklist


happy-san:
@krunal1313 Few nitpicks:

-  Instead of the `curation_timeout` which was inadvertently changed by Harisaran's curation PR, it should be `override_timeout` i.e. if we want to ignore the search cutoff timeout or not. I think replacing all instances of `curation` with `override` in `filter_result_iterator` header and cpp file should be fine.

- Also, we could rename `is_metadata_free_subtree()` to something like `subtree_contains_refence_filter()` since we're only checking `filter_exp.referenced_collection_name.empty()`

krunal1313:
@happy-san I've updated the PR.

